Issue #493, opened 2021-11-06
Add distribution support for common package managers


ezekg:
Related to #478, #489, #490, #491. Foundation implemented in #679.

Add unique subdomain for each engine, e.g. Composer `https://composer.pkg.keygen.sh`.

- [ ] `sparkle`
- [ ] `squirrel`
- [ ] `electron` (alias for `squirrel`)
- [x] `tauri`: #753 and #860
- [x] `npm`: #489
- [ ] `unity` (alias for `npm`?)
- [x] `gems`: #491
- [ ] `composer` ([requires subdomains](https://github.com/keygen-sh/keygen-api/issues/490#issuecomment-960394554))
- [x] `docker` (OCI): #911
- [x] `pypi`: #679
- [ ] `wp`: https://github.com/keygen-sh/keygen-api/issues/904
- [ ] `homebrew`
- [x] `raw`: #892

Issue #526, opened 2021-12-16
Licenses should have many products


ezekg:
One license to rule them all. Policies should also not be tied to a particular product.

ezekg:
Is this even needed anymore with [the introduction of packages](#679)?

Issue #521, opened 2021-12-08
Add product plugin resource


ezekg:
Share a license across a product with multiple plugins. Manage releases per-plugin separately from the product’s releases.

```rb
class ProductPlugin
  belongs_to :product
end

class Product
  has_many :plugins,
    class_name: 'ProductPlugin'
end

class Release
  belongs_to :plugin,
    class_name: 'ProductPlugin',
    optional: true
end

class LicenseProductPlugin
  belongs_to :product_plugin
  belongs_to :license
end

class License
  has_many :license_product_plugins
  has_many :plugins,
    through: :license_product_plugins,
    source: :product_plugin
end
```

Should licenses manage their accessible plugins separately from the product’s?

ezekg:
Could also call these "packages" instead of "plugins."

PR #679, opened 2022-11-14
Add PyPI package registry

ezekg:
Part of #493. Implements [PyPI's simple package repository](https://packaging.python.org/en/latest/guides/hosting-your-own-index/). Adds release engines and release packages. And as a side-effect, adds support for non-JSON API mime types.

```bash
# Install "example" package from Keygen (other packages are still redirected to PyPI)
pip install example \
  --index-url 'https://api.keygen.sh/v1/accounts/demo/engines/pypi/simple'
```

## Post-deploy

```bash
# Add new permissions and event types
rake db:seed

# Migrate existing role permissions
rake keygen:permissions:admins:add[package.create,package.delete,package.read,package.update,engine.read]
rake keygen:permissions:environments:add[package.create,package.delete,package.read,package.update,engine.read]
rake keygen:permissions:products:add[package.create,package.delete,package.read,package.update,engine.read]
rake keygen:permissions:licenses:add[package.read,engine.read]
rake keygen:permissions:users:add[package.read,engine.read]
```

Ref: [PEP 503](https://peps.python.org/pep-0503/).

### DIFF
--- a/app/models/release_package.rb
+++ b/app/models/release_package.rb
@@ -14,6 +14,7 @@ class ReleasePackage < ApplicationRecord
   belongs_to :engine,
     foreign_key: :release_engine_id,
     class_name: 'ReleaseEngine',
+    autosave: true,
     optional: true
   has_many :releases,
     inverse_of: :package,


app/controllers/api/v1/release_packages/pypi/simple_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    private
+
+    attr_reader :product
+
+    def set_product
+      scoped_products = authorized_scope(current_account.products)
+
+      # TODO(ezekg) Add a distribution_engine attribute to product?
```
ezekg, line 34: I'm thinking we should add a `distribution_engine` to products, releases and artifacts. That way we can assert we only pull in artifacts that support PyPI, while allowing products to have various engines if needed. We can have releases inherit their engine from its product by default, and artifacts from their release.

app/views/api/v1/packages/pypi/simple/index.html.haml
```diff
@@ -0,0 +1,12 @@
+-# frozen_string_literal: true
+!!!
+%html
+  %head
+    %title Links for #{product.name} by #{account.name}
+  %body
+    %h1 Links for #{product.name} by #{account.name}
+
+    - artifacts.each do |artifact|
```
ezekg, line 9: Add caching?

db/schema.rb
```diff
@@ -409,13 +409,13 @@
     t.uuid "account_id"
     t.string "url"
     t.string "distribution_strategy"
-    t.string "slug"
```
ezekg, line 412: This somehow snuck in in #687.

app/controllers/concerns/rendering.rb
```diff
@@ -0,0 +1,46 @@
...
+
+          render html: Rack::Utils::HTTP_STATUS_CODES[code],
+                 status:
+        end
+      end
+    end
+
+    private
+
+    def set_content_type = response.headers['Content-Type'] = mime_type
```
ezekg, line 40: This seems dirty. Should be handled by the controller's format.

app/controllers/concerns/rendering.rb
```diff
@@ -0,0 +1,46 @@
...
+    # need to provide detailed errors for e.g. PyPI.
+    Rack::Utils::SYMBOL_TO_STATUS_CODE.each do |status, code|
+      if Rack::Utils::STATUS_WITH_NO_ENTITY_BODY.include?(code)
+        define_method :"render_#{status}" do |*, **|
+          skip_verify_authorized!
+
+          head status
+        end
+      else
+        define_method :"render_#{status}" do |*, **|
```
ezekg, line 29: This seems wrong. Ideally, these render methods should `respond_to` our supported formats, including HTML.

app/controllers/concerns/default_headers.rb
```diff
@@ -136,20 +127,4 @@ def add_version_header
   def add_powered_by_header
     response.headers['X-Powered-By'] = 'Ruby, Rails, and a lot of coffee. (And the occasional Islay.)'
   end
-
-  def artifact_route?
```
ezekg, line 140: Bye bye 👏

config/application.rb
```diff
@@ -82,6 +82,15 @@ class Application < Rails::Application
     # Use Sidekiq for background jobs
     config.active_job.queue_adapter = :sidekiq
 
+    # Include all helpers
+    config.action_controller.include_all_helpers = true
+
+    # Default URL options (namely host)
+    Rails.application.default_url_options = {
```
ezekg, line 89: I think this should be `application.routes.default_url_options`

ezekg, line 89:
```rb
config.action_dispatch.return_only_media_type_on_content_type = true
```

ezekg: Worth mentioning: https://github.com/rails/rails/pull/47758

app/controllers/concerns/default_headers.rb
```diff
@@ -100,15 +72,34 @@ def add_rate_limiting_headers
...
+
+    # We consider both application/vnd.api+json and application/json as
+    # synonyms of the :jsonapi format, but we still want to respond
+    # with application/json if asked to do so, for compatibility
+    # with various HTTP clients, e.g. cpp-rest-sdk.
+    jsonapi = Mime::Type.lookup_by_extension(:jsonapi).to_s
+    json    = Mime::Type.lookup_by_extension(:json).to_s
+    accept  = request.headers['Accept']
+
+    if json.in?(accept)
```
ezekg, line 90: Use `request.content_type` to retain charset.

config/application.rb
ezekg: Tests should be updated to allow a charset.

app/models/product.rb
```diff
@@ -100,12 +100,20 @@ class Product < ApplicationRecord
 
   scope :for_license, -> id {
     joins(:licenses).where(licenses: { id: })
+                    .licensed
                     .distinct
+                    .union(
+                      self.open,
```
ezekg, line 106: Update product authz tests.

config/routes.rb
```diff
@@ -75,311 +75,322 @@
...
 
-    resources :analytics, only: [] do
-      collection do
-        scope :actions, module: 'analytics/actions' do
-          get :top_licenses_by_volume, path: 'top-licenses-by-volume', to: 'counts#top_licenses_by_volume'
-          get :top_urls_by_volume,     path: 'top-urls-by-volume',     to: 'counts#top_urls_by_volume'
-          get :top_ips_by_volume,      path: 'top-ips-by-volume',      to: 'counts#top_ips_by_volume'
-          get :count,                                                  to: 'counts#count'
-        end
+    namespace :release_packages, path: 'packages' do
```
ezekg, line 377: Rename is `registries` so we can use `packages` later? E.g. in #521.

ezekg: Or use `pkg` subdomain?

db/migrate/20230712192044_add_distribution_engine_to_products.rb
```diff
@@ -0,0 +1,7 @@
+class AddDistributionEngineToProducts < ActiveRecord::Migration[7.0]
+  def change
+    add_column :products, :distribution_engine, :string, null: true
```
ezekg, line 3: Should we add `engine` to releases and artifacts? May be easier to just leave it alone for now.

features/api/v1/packages/pypi/simple.feature
```diff
@@ -0,0 +1,258 @@
...
+      """
+      {
+        "distributionEngine": "PYPI",
+        "code": "package1"
+      }
+      """
+    And I am an admin of account "test1"
+    And I use an authentication token
+    When I send a GET request to "/accounts/test1/packages/pypi/simple/package1"
+    Then the response status should be "200"
```
ezekg, line 56: Need to run at least 1 assertion on the response body, asserting that it's HTML.

Ideally, we'd assert that _n_ artifacts are present.

db/migrate/20230712192044_add_distribution_engine_to_products.rb
ezekg: Right now, it doesn't make sense to change an engine once set, so let's just leave it as-is right now. You can't license multiple engines under the same product yet, anyways, so we can revisit.

features/api/v1/packages/pypi/simple.feature
ezekg:
```gherkin
And the response body should be an HTML document with the following markup:
      """
      <a href="https://api.keygen:sh/v1/accounts/foo/artifacts/bar">...</a>
      """
```

ezekg: Or we could use Nokogiri:

```gherkin
And the response body should be an HTML document with the following xpaths:
  """
  /html/body/a[@href="https://api.keygen:sh/v1/accounts/foo/artifacts/bar"]
  """
  ```

File: app/views/api/v1/release_packages/pypi/simple/index.html.haml [hunk 13]
```diff
@@ -7,6 +7,6 @@
     %h1 Links for #{product.name} by #{account.name}
 
     - artifacts.each do |artifact|
-      - url  = url_for([:v1, account, artifact, anchor: checksum_for(artifact, delimiter: '=')])
+      - url = url_for([:v1, account, artifact, anchor: checksum_for(artifact, delimiter: '='), protocol: 'https'])
```
ezekg, line 10: I don't understand why the default URL options in `config/application.rb` are not applying here...

config/routes.rb
ezekg: Meh. I think this is fine. We should prefer #526 instead, anyways.

config/routes.rb
```diff
@@ -75,311 +75,332 @@
...
-          get :top_licenses_by_volume, path: 'top-licenses-by-volume', to: 'counts#top_licenses_by_volume'
-          get :top_urls_by_volume,     path: 'top-urls-by-volume',     to: 'counts#top_urls_by_volume'
-          get :top_ips_by_volume,      path: 'top-ips-by-volume',      to: 'counts#top_ips_by_volume'
-          get :count,                                                  to: 'counts#count'
-        end
+    # Release packages can support and respond with a variety of mime types, so
+    # we're defining those routes here with their own unique constraints.
+    namespace :release_packages, path: 'packages' do
+      scope :pypi, module: :pypi, constraints: MimeTypeConstraint.new(:html, raise_on_no_match: true), defaults: { format: :html } do
+        get 'simple/:id', to: 'simple#index', as: :pypi_packages
```
ezekg, line 389: We also need a `get 'simple'` route [according to PEP 503](https://peps.python.org/pep-0503/). So rename the current action to `#show` and create an `#index` listing all products with a `PYPI` distribution strategy.

ezekg: Note that all links to HTML pages, i.e. in the new `#index` action, should have a trailing slash according to the spec.

We can use `url_for(..., trailing_slash: true)`.

app/policies/release_packages/pypi/simple/product_policy.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module ReleasePackages::Pypi::Simple
+  class ProductPolicy < ApplicationPolicy
+    skip_pre_check :verify_authenticated!
+
+    def index?
```
ezekg, line 7: Add specs.

app/policies/release_packages/pypi/simple/release_artifact_policy.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+module ReleasePackages::Pypi::Simple
+  class ReleaseArtifactPolicy < ApplicationPolicy
+    skip_pre_check :verify_authenticated!
+
+    def show?
```
ezekg, line 7: Add specs.

app/models/permission.rb
```diff
@@ -99,6 +99,8 @@ class Permission < ApplicationRecord
 
     metric.read
 
+    package.read
```
ezekg, line 102: Good time to put together migrations for adding a new permission to all roles. Right now, we only have a Rake task for adding new permissions to admins. We'll need to add tasks for the other roles as well.

app/views/api/v1/release_packages/pypi/simple/index.html.haml
```diff
@@ -0,0 +1,10 @@
+-# frozen_string_literal: true
+- title = "Packages for #{account.name}"
+- content_for(:title) { title }
+
+%h1= title
+
+- packages.each do |package|
+  - url = url_for([:v1, account, :release_packages, :pypi_simple_package, id: package.id, format: nil, trailing_slash: true])
+
+  = link_to(package.code, url, data: package.metadata)
```
ezekg, line 10: Should we require that all products have a code, or write a script to set a default?

```ruby
Product.find_each do |product|
  product.update!(code: product.name.parameterize)
end
```

app/controllers/api/v1/release_packages/pypi/simple_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def show
+      product = FindByAliasService.call(
+        authorized_scope(current_account.products.pypi),
+        id: params[:id],
+        aliases: :code,
+      )
+
+      artifacts = authorized_scope(apply_scopes(product.release_artifacts.pypi)).limit(1_000)
```
ezekg, line 32: Should this be `order_by_version`?

ezekg, line 32: Is a limit of `1_000` high enough? Looking at projects like Rails and FastAPI, it's big enough, but projects like React have > 1,000 versions. We can always change this, of course. Just something to note.

ezekg, line 32: We should `preload(:product, :release)`.

app/views/api/v1/release_packages/pypi/simple/index.html.haml
ezekg: A simple way to do this would be to insert a `package.read` permission for every role and token that has a `artifact.read` permission.

app/models/permission.rb
ezekg: A simple way to do this would be to insert a `package.read` permission for every role and token that has a `artifact.read` permission.

app/views/api/v1/release_packages/pypi/simple/index.html.haml
ezekg: I feel like we should do both. PEP 503 requirs that this match the package name.

app/models/release_package.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+class ReleasePackage
+  include ActiveModel::Model
```
ezekg, line 4: Let's turn this into an active record. We can replace the product's `distribution_engine` with a package `engine`, and allow releases to be associated to a specific package, allowing multiple distributions under the same product.

app/views/api/v1/release_packages/pypi/simple/index.html.haml
ezekg: We should replace a product's `code` with a `code` column on the new package active record.

config/initializers/action_policy.rb
```diff
@@ -7,5 +7,15 @@
   config.auto_inject_into_channel          = false
 end
 
+# Lookup chain should support arrays of like-activemodels.
+ActionPolicy::LookupChain.chain << -> records, **options {
+  case records
+  in Array(_ => first, *) if records.all? { _1.class == first.class }
```
ezekg, line 13: Should this be `_1.policy_class == first.policy_class`?

app/models/release_package.rb
ezekg:
```rb
class ReleasePackage
  attribute :product_id, :uuid
  attribute :keywords, :string, array: true
  attribute :author, :string
  attribute :license, :string
  attribute :engine, :string
end
```

app/models/release.rb
```diff
@@ -139,6 +151,16 @@ class Release < ApplicationRecord
       if: :tag?,
     }
 
+  # Assert that package matches the release's product.
+  validate on: %i[create update] do
+    next unless
+      release_package_id_changed?
+
+    unless package.nil? || package.product_id == product_id
+      errors.add :package, :not_allowed, message: 'package product must match release product'
```
ezekg, line 160: I don't like the denormalization here, but there's really no way around it since package could be nil (meaning we can't have the release associated with a product through the package).

ezekg: It's essentially the same thing as we have for `account_id` and `environment_id`, so meh.

features/step_definitions/resource_steps.rb
```diff
@@ -175,9 +182,22 @@
...
+  associated_records =
+      case association.underscore.pluralize
+      when 'processes'
+        @account.machine_processes
+      when 'artifacts'
+        @account.release_artifacts
+      when 'packages'
+        @account.release_packages
+      when 'engines'
+        @account.release_engines
```
ezekg, line 195: If these associations were named correctly on the account, this wouldn't be needed.

app/controllers/api/v1/release_packages_controller.rb
```diff
@@ -0,0 +1,139 @@
...
+      format :jsonapi
+
+      param :data, type: :hash do
+        param :type, type: :string, inclusion: { in: %w[package packages] }
+        param :attributes, type: :hash do
+          param :name, type: :string
+          param :key, type: :string
+          param :engine, type: :string, inclusion: { in: %w[pypi] }, optional: true, allow_nil: true,
+            transform: -> (_, key) {
+              [:engine_attributes, { key: }]
```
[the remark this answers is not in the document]
ezekg: Added test

app/controllers/application_controller.rb
```diff
@@ -68,121 +73,184 @@ def render_no_content(**kwargs)
...
-    render status: :forbidden, json: {
-      meta: { id: request.request_id },
-      errors: [{
-        title: "Access denied",
-        detail: "You do not have permission to complete the request",
-        **kwargs,
-      }],
-    }
+    respond_to do |format|
+      format.any(:jsonapi, :json, :binary) {
```
ezekg, line 77: Is there a better way to do this? Essentially, I want `format.any(except: :html)`.

app/controllers/application_controller.rb
```diff
@@ -4,8 +4,9 @@ class ApplicationController < ActionController::API
   # FIXME(ezekg) Why is this needed?
   self.default_url_options = Rails.application.default_url_options
 
-  include ActionController::MimeResponds
   include Rendering::JSON
+  using Rendering::Refinements
```
ezekg, line 8: Including this in the concern itself doesn't work, so gotta do it here.

lib/tasks/keygen/permissions.rake
```diff
@@ -3,47 +3,124 @@
...
+    task :add, %i[type] => %i[environment] do |_, args|
+      batch_size = ENV.fetch('BATCH_SIZE') { 1_000 }.to_i
+      batch_wait = ENV.fetch('BATCH_WAIT') { 0.1 }.to_f
+
+      model = args[:type].to_s.classify.safe_constantize
+
+      # Split args up into ID and permission buckets
+      ids,
+      permissions,
+      * = args.extras.chunk   { Permission::ALL_PERMISSIONS.include?(_1) }
```
ezekg, line 15: This needs to be sorted, no? That way we don't mix up the ID and permission arrays.

lib/tasks/keygen/permissions.rake
```diff
@@ -3,47 +3,124 @@
...
   namespace :permissions do
+    task :add, %i[type] => %i[environment] do |_, args|
+      batch_size = ENV.fetch('BATCH_SIZE') { 1_000 }.to_i
+      batch_wait = ENV.fetch('BATCH_WAIT') { 0.1 }.to_f
+
+      model = args[:type].to_s.classify.safe_constantize
+
+      # Split args up into ID and permission buckets
+      ids,
+      permissions,
```
ezekg, line 14: Rename this to `new_permissions`?

lib/tasks/keygen/permissions.rake
```diff
@@ -3,47 +3,124 @@
...
+      * = args.extras.chunk   { Permission::ALL_PERMISSIONS.include?(_1) }
+                     .collect { _2 }
+
+      records = model.includes(:account, role: { role_permissions: :permission })
+                     .where(id: ids)
+
+      records.find_each(batch_size:) do |record, i|
```
ezekg, line 21: Drop unused index

Issue #569, opened 2022-01-22
Add expiration strategies for allowing minor/patch upgrades


ezekg:
For a perpetual fallback license, allowing users with an expired license to still obtain access to patch releases would be beneficial.

ezekg:
The `MAINTAIN_ACCESS` expiration strategy is good enough here.

PR #734, opened 2023-06-13
Add last validated version and checksum

### LINKED ISSUE #620: Track a license's current release version
I believe we already have this information, but it may not be easily queryable. We could query a license's `release.downloaded` and `release.upgraded` event logs (and for its user), and then extract its highest `metadata->next` or `metadata->version` to determine which version they're currently on.

But this doesn't really work with `OPEN` release distributions… and it seems brittle because it relies on download events, not on usage events (e.g. a license could download a particular version but never use it).

ezekg:
Closes #620. Foundation for #569.

### DIFF
--- a/app/serializers/license_serializer.rb
+++ b/app/serializers/license_serializer.rb
@@ -28,6 +28,9 @@ class LicenseSerializer < BaseSerializer
   attribute :protected do
     @object.protected?
   end
+  attribute :version do
+    @object.last_validated_version
+  end
   attribute :max_machines
   attribute :max_processes
   attribute :max_cores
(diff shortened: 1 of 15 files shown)


app/serializers/license_serializer.rb
```diff
@@ -28,6 +28,9 @@ class LicenseSerializer < BaseSerializer
   attribute :protected do
     @object.protected?
   end
+  attribute :version do
+    @object.last_validated_version
```
ezekg, line 32: Would be cool to track this over time, and not just the version for the most recent scoped validation.

But we can do that later.

PR #738, opened 2023-06-23
Refactor pattern matching for role checks

ezekg:
Cleans up role checks to use a typed `Role(...)` pattern and symbol role names.


ezekg:
Missed: https://github.com/keygen-sh/keygen-api/blob/master/app/models/user.rb#L44

PR #742, opened 2023-06-27
Fix issue where dead machines/processes would never be culled

### LINKED ISSUE #739: Add periodic job to clean up orphaned dead machines and processes
When a policy's heartbeat duration is changed while there are heartbeat monitors in-progress, some machine (and process) heartbeat monitors may become out-of-date and the dead will eventually be orphaned if no further pings are sent.

For example:

Given a policy with a heartbeat duration of 600, if a machine is activated and a heartbeat monitor is started, and if the policy is then changed to have a heartbeat duration of 86000, when the machine's monitor runs at 600 seconds from time of activation, it will now see that the machine is not dead and will send a pong event and exit. Eventually, this machine dies if no further pings are sent, and since no pings were sent, there is then no heartbeat monitor to handle its death, and it remains unculled.

The same could happen for processes. I'm sure there are other scenarios like this. So we really need to set up a periodic job that deactivates orphaned dead machines i.e. those that no longer have a heartbeat monitor attached.

ezekg:
Closes #739.


ezekg:
I'd like to add logging for Sidekiq jobs and their arguments, to make searching and debugging these types of jobs easier.

We should create an opt-in middleware for logging a job's arguments.

app/workers/machine_heartbeat_worker.rb
```diff
@@ -53,15 +53,18 @@ def perform(machine_id)
       account: machine.account,
       resource: machine,
     )
+
+    # The machine no longer has a heartbeat monitor (until next ping)
+    machine.update(
```
ezekg, line 58: Should this assert that the current JID is still correct?

```ruby
Machine.where(account_id: machine.account_id, id: machine.id, heartbeat_jid: jid).limit(1)
       .update(heartbeat_jid: nil)
```

PR #747, opened 2023-07-05
Add UUID type for typed_params

### LINKED ISSUE #745: Add uuid param type for relationship IDs
It's a common misunderstanding when a user tries to e.g. create a machine for a license using the license's key instead of its UUID.

To communicate better errors, we should add a `:uuid` type to our `typed_params` initializer and update all relationship params to use `:uuid`:

```diff
param :relationships do
  param :license do
    param :data do
-     param :id, type: :string
+     param :id, type: :uuid
      param :type, type: :string
    end
  end
end
```

ezekg:
Closes #745. Saves a few database hits, and reduces the confusion outlined in the aforementioned issue.

### DIFF
--- a/config/initializers/typed_params.rb
+++ b/config/initializers/typed_params.rb
@@ -24,3 +24,11 @@ TypedParams.types.register(:metadata,
     }
   },
 )
+
+TypedParams.types.register(:uuid,
+  archetype: :string,
+  name: 'UUID',
+  match: -> value {
+    value.is_a?(String) && UUID_RE.match?(value)
+  },
+)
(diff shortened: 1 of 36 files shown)


config/initializers/typed_params.rb
```diff
@@ -24,3 +24,11 @@
     }
   },
 )
+
+TypedParams.types.register(:uuid,
+  archetype: :string,
```
ezekg, line 29: I don't really like that this has an archetype, but [`typed_params` currently appends new types](https://github.com/keygen-sh/typed_params/blob/7d95eb2d7a6e311158565c9931191e6b24bd4de1/lib/typed_params/types.rb#L16), so matching against a type that is also another type will always match the former type (in this case, string). Maybe we should add a `:prepend` option to `#register`, or add a `#prepend` method?

Either way, not gonna lose sleep over it. But may be something to eventually look at.

Issue #390, opened 2020-12-14
Add components relationship to machines


ezekg:
Related to #160. Looking for a nice way to support multiple fingerprints per-machine, e.g. MOBO ID, HDD ID, MAC addresses, etc., which we can use along with #387's `fingerprint_matching_strategy`.

ezekg:
May need to add new matching strategies specific to components, to e.g. enforce that all fingerprints belong to the same machine, since right now matching of the `fingerprints` validation scope is cross-machine.

ezekg:
May also need to add a `UNIQUE_PER_MACHINE` fingerprint uniqueness strategy, to allow the same component fingerprint across multiple machines. Would be needed in environments where machines use the same underlying hardware, e.g. cloud.

PR #755, opened 2023-08-25
Add embedded machine components

ezekg:
Follow up to #390. Allows components to be added during machine creation, saving on superfluous requests.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -498,6 +498,8 @@ class ApplicationController < ActionController::API
     render_unprocessable_resource e.record
   rescue ActiveRecord::RecordNotUnique
     render_conflict # Race condition on unique index
+  rescue ActiveRecord::NestedAttributes::TooManyRecords
+    render_unprocessable_entity detail: 'too many records'
   rescue ActiveModel::ValidationError => e
     render_unprocessable_resource e.model
   rescue Encoding::CompatibilityError,
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -264,17 +264,16 @@ class ApplicationController < ActionController::API
         klass   = resource.class
         pointer = nil
 
-        if klass.respond_to?(:reflect_on_association) &&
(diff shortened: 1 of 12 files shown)


app/models/machine_component.rb
```diff
@@ -34,15 +34,32 @@ class MachineComponent < ApplicationRecord
...
+    # Special case where fingerprints could be duplicated in the actual nested
+    # association params, so this adds better error messaging vs a plain
+    # 409 Conflict error via the unique index violation.
+    if !machine.persisted? && machine.components_attributes_assigned?
+      components = machine.components
+
+      component.errors.add(:fingerprint, :conflict, message: 'is duplicated') if
+        components.reject { _1 == component }
+                  .find { _1.fingerprint == component.fingerprint }
+                  .present?
```
ezekg, line 46: Is there a cleaner way to check for duplicates in Ruby/Rails?

app/controllers/application_controller.rb
```diff
@@ -264,17 +264,16 @@ def render_unprocessable_resource(resource)
...
-          when err == "must exist" && src.size > 1
-            src.insert 1, :data
-            src.insert -2, :relationships
-          when err != "must exist" && src.size > 1
-            src.insert 1, :data
-            src.insert -2, :attributes
+        if path.first != "role" && klass.respond_to?(:reflect_on_association) &&
+           (assoc = klass.reflect_on_association(path.first))
+          if src.size > 1
+            if assoc.klass&.reflect_on_association(path.last)
```
ezekg, line 270: Refactor this. Code quality sucks.

app/models/machine.rb
```diff
@@ -34,6 +35,9 @@ class ResurrectionExpiredError < StandardError; end
 
   has_environment default: -> { license&.environment_id }
 
+  accepts_nested_attributes_for :components, limit: 20, reject_if: :reject_associated_records_for_components
```
ezekg, line 38: We need to catch the limit error, `ActiveRecord::NestedAttributes::TooManyRecords`.

PR #756, opened 2023-08-30
Refactor unprocessable resource renderer

ezekg:
Refactors the dragons into something that doesn't suck (as much).

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -256,9 +256,51 @@ class ApplicationController < ActionController::API
     errors = resource.errors.as_jsonapi
     meta   = { id: request.request_id }
 
-    # FIXME(ezekg) Move all 'custom' logic out of .to_jsonapi and handle
-    #              each individual error code explicitly, rewriting by
-    #              pointer if needed. Also add links explicitly.
+    errors.each do |error|
+      # Fixup various error codes and pointers to match our objects, e.g.
+      # some relationships are invisible, exposed as attributes.
+      case error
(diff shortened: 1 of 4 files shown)


lib/keygen/jsonapi/errors.rb
```diff
@@ -0,0 +1,111 @@
...
+                  pointer << :id
+                else
+                  pointer << :attributes << source
+                end
+              end
+
+              # Simplify and clarify validation error codes i.e. we don't need
+              # to expose our validators to the world.
+              prefix = source == :base ? @base.class.name.underscore : path.select { _1 in Symbol }.join('_')
+              code   = case error.type
```
ezekg, line 76: I don't love this (would like to move it to application controller), but meh.

app/controllers/application_controller.rb
```diff
@@ -253,149 +253,66 @@ def render_service_unavailable(**kwargs)
...
+    errors.each do |error|
+      # Fixup various error codes and pointers to match our objects, e.g.
+      # some relationships are invisible, exposed as attributes.
+      case error
+      in source: { pointer: %r{^/data/relationships/users/(.*+)} } if resource in Account
+        error.pointer = "/data/relationships/admins/#{$1}"
+      in source: { pointer: %r{^/data/attributes/permission_?ids$}i },
+         code: /^PERMISSION_IDS_(.+)/
+        error.pointer = '/data/attributes/permissions'
+        error.code    = "PERMISSIONS_#{$1}"
```
ezekg, line 268: Order dependent 🙈

ezekg: Let's see how far this takes us

Issue #628, opened 2022-05-04
Add Keygen-* request/response headers to request logs


ezekg:
E.g. `Keygen-Signature`, `Keygen-Version`, etc. Add 2 new `jsonb` columns, `request_logs.request_headers` and `request_logs.response_headers`.

ezekg:
Related to #519.

ashutosh887:
Let me try this @ezekg

ezekg:
@ashutosh887 go ahead.

PR #760, opened 2023-09-12
Add headers and timings to request logs

### LINKED ISSUE #519: Add request timing and headers to request logs
E.g. `Content-Type`, `Content-Length` and `Accept` request headers, and `Content-Type` and `Content-Length` response headers. The `Authorization` header MUST NOT be included, of course. Timings may be able to be derived from the `X-Request-Start` header or something from Rails' internals (since Rails internally keeps timings on various things).

Add 2 new `jsonb` columns, `request_logs.request_headers` and `request_logs.response_headers`. Add the additional arguments to the request log Sidekiq job, taking consideration into the process [for safely changing a job's arguments](https://keygen.sh/blog/how-to-change-the-signature-of-a-sidekiq-job/).

Ref: https://github.com/discourse/discourse/blob/1472e47aae5bfdfb6fd9abfe89beb186c751f514/lib/middleware/request_tracker.rb

ezekg:
Closes #519 and #628. This should help with debugging a number of requests, and also provide visibility into timings.

### DIFF
--- a/app/controllers/concerns/default_headers.rb
+++ b/app/controllers/concerns/default_headers.rb
@@ -3,15 +3,9 @@
 module DefaultHeaders
   extend ActiveSupport::Concern
 
-  include CurrentAccountScope
-  include SignatureHeaders
   include RateLimiting
 
   included do
-    # NOTE(ezekg Run signature header validations after current account has been set, but
-    #            before the controller action is processed.
-    after_current_account :validate_accept_signature_header!
-
     # NOTE(ezekg) We're using an *around* action here to ensure these headers are always
     #             sent, even when an error has halted the action chain.
     around_action :add_default_headers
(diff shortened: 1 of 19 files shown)


app/controllers/concerns/default_headers.rb
```diff
@@ -103,9 +103,9 @@ def add_content_type_header
   end
 
   def add_whoami_headers
-    response.headers['Keygen-Account-Id'] = current_account&.id
-    response.headers['Keygen-Bearer-Id']  = current_bearer&.id
-    response.headers['Keygen-Token-Id']   = current_token&.id
+    response.headers['Keygen-Account'] = current_account&.id
+    response.headers['Keygen-Bearer']  = current_bearer&.id
+    response.headers['Keygen-Token']   = current_token&.id
```
ezekg, line 108: Breaking change. Revert or add a migration.

ezekg: Not worth a migration, IMO, so just revert.

PR #764, opened 2023-10-10
Upgrade to Rails 7.1

### LINKED ISSUE #615: Migrate from secrets.yml to credentials.yml
The former was deprecated in Rails 5.2.

ezekg:
Closes #615. Upgrades to Rails 7.1. Upgrade docs can be found [here](https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-7-0-to-rails-7-1).


ezekg:
Need to drop usage of `secrets.yml` in favor of env vars.

Gemfile
```diff
@@ -52,8 +50,8 @@ gem 'kaminari', '~> 1.2.0'
 
 # Postgres/DB extensions
 gem 'active_record_union'
-gem 'active_record_distinct_on'
-gem 'activerecord_where_assoc'
+gem 'active_record_distinct_on', github: 'ezekg/active_record_distinct_on', branch: 'feature/support-rails-7-1'
+gem 'activerecord_where_assoc', github: 'ezekg/activerecord_where_assoc'
```
ezekg, line 54: https://github.com/MaxLap/activerecord_where_assoc/issues/13

Gemfile
```diff
@@ -52,8 +50,8 @@ gem 'kaminari', '~> 1.2.0'
 
 # Postgres/DB extensions
 gem 'active_record_union'
-gem 'active_record_distinct_on'
-gem 'activerecord_where_assoc'
+gem 'active_record_distinct_on', github: 'ezekg/active_record_distinct_on', branch: 'feature/support-rails-7-1'
```
ezekg, line 53: https://github.com/alecdotninja/active_record_distinct_on/pull/28

ezekg:
New active record encryption scheme is breaking with old values, so need to figure it out. Either non-deterministic values fail to decrypt, or deterministic values fail. I can't find a way to make both succeed and need to spend a bit more time debugging.

Ref: https://github.com/rails/rails/issues/48204

config/application.rb
```diff
@@ -64,12 +64,14 @@ class Application < Rails::Application
     config.active_record.encryption.key_derivation_salt = ENV.fetch('ENCRYPTION_KEY_DERIVATION_SALT')
 
     # TODO(ezekg) Remove these once our data is migrated
+    config.active_record.encryption.hash_digest_class                             = OpenSSL::Digest::SHA256
+    config.active_record.encryption.support_sha1_for_non_deterministic_encryption = true
```
ezekg, line 68: Drop this

config/application.rb
```diff
@@ -64,12 +64,14 @@ class Application < Rails::Application
...
+    config.active_record.encryption.store_key_references     = true
```
ezekg, line 69: Remove this.

ezekg:
Still sitting on this due to https://github.com/rails/rails/issues/48204.

app/models/machine_process.rb
```diff
@@ -90,21 +90,21 @@ class ResurrectionExpiredError < StandardError; end
...
   end
 
-  scope :for_product, -> product { joins(license: :product).where(products: product) }
-  scope :for_license, -> license { joins(:license).where(licenses: license) }
-  scope :for_machine, -> machine { where(machine: machine) }
-  scope :for_user,    -> id { joins(:license).where(licenses: { user_id: id }) }
+  scope :for_product, -> product { joins(license: :product).where(product: { id: product }) }
+  scope :for_license, -> license { joins(:license).where(license: { id: license }) }
+  scope :for_machine, -> machine { where(machine:) }
+  scope :for_user,    -> id { joins(:license).where(license: { user_id: id }) }
```
ezekg, line 96: Drop these modifications. They're unrelated to the upgrade and are going to cause conflicts with #774.

ezekg: Apparently they're not unrelated...

ezekg: What is Rails doing here that causes `where(foo:)` to error while `where(foo: { id: })` doesn't? I'd expect these to be synonymous.

ezekg: It's because in this case, these model scopes can be called from controller scopes, and the controller provides IDs, not model records, so we need to query by `id:` to support both ID and record queries.

ezekg: https://github.com/rails/rails/pull/50829

ezekg:
I'd like to get this out before #774.

spec/shared/encryptable.rb
```diff
@@ -0,0 +1,53 @@
...
+                {
+                  id: record.id,
+                  value: enc,
+                },
+              ],
+            ),
+          )
+
+          expect { record.reload.read_attribute(encrypted_attribute) }.to_not raise_error
+          expect(record.read_attribute(encrypted_attribute) ).to eq dec
```
ezekg, line 48: Literally wrote this for my own sanity. So much conflicting info in https://github.com/rails/rails/issues/48204.

config/application.rb
ezekg: Shouldn't have added this. Broke license key lookups, since the lookups were now including the key reference. 🤦‍♂️

PR #766, opened 2023-10-13
Update event logs pruning to only prune duplicates

### LINKED ISSUE #762: Only duplicate high-volume events per-day should be pruned
Right now, we prune all high-volume event logs after 90 days. But I'd like to add a bit of nuance here so we don't lose license usage data. I.e. if a license is validated 1,000 times per day, we should at least keep a record of the fact that it was validated that day. Best way to do this, I think, would be to remove duplicate events per resource.

ezekg:
Closes #762. Keeps latest event log per-resource, per-day and per-event, removing all duplicates for space savings.

### DIFF
--- a/app/workers/prune_event_logs_worker.rb
+++ b/app/workers/prune_event_logs_worker.rb
@@ -50,9 +50,18 @@ class PruneEventLogsWorker < BaseWorker
                             .where('created_at < ?', BACKLOG_DAYS.days.ago.beginning_of_day)
                             .where(event_type_id: event_type_ids)
 
+        # Keep the latest log per-resource for each day and event type (i.e. discard duplicates)
+        kept_logs = event_logs.distinct_on(:resource_id, :resource_type, :event_type_id, :created_date)
+                              .reorder(:resource_id, :resource_type, :event_type_id,
+                                created_date: :desc,
+                              )
+                              .select(:id)
+
         batch += 1
         count = event_logs.limit(BATCH_SIZE)
-                          .delete_all
+                          .delete_by(
+                            "id NOT IN (#{kept_logs.to_sql})",
+                          )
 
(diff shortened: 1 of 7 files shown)


app/workers/prune_event_logs_worker.rb
```diff
@@ -50,9 +50,18 @@ def perform
                             .where('created_at < ?', BACKLOG_DAYS.days.ago.beginning_of_day)
                             .where(event_type_id: event_type_ids)
 
+        # Keep the latest log per-resource for each day and event type (i.e. discard duplicates)
+        kept_logs = event_logs.distinct_on(:resource_id, :resource_type, :event_type_id, :created_date)
```
ezekg, line 54: Adding a composite index on these columns may help improve performance here. Now sure if the cost for writes is worth it, though, since we do have an index on each of these columns. We'll have to look at real-world performance after we deploy.

db/migrate/20231013164654_add_created_date_to_event_logs.rb
```diff
@@ -0,0 +1,11 @@
+class AddCreatedDateToEventLogs < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_column :event_logs, :created_date, :date, null: true
```
ezekg, line 5: Need to eventually make this non-null.

spec/workers/prune_event_logs_worker_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+  # See: https://github.com/mhenrixon/sidekiq-unique-jobs#testing
+  before do
+    Sidekiq::Testing.inline!
+  end
+
+  after do
+    Sidekiq::Worker.clear_all
+  end
+
+  it 'should only prune high-volume event logs' do
```
ezekg, line 19: Remove "only"

PR #768, opened 2023-11-09
Update license user association from has one to has many

### LINKED ISSUE #534: A license should have multiple users
Related to #533. For project-based licenses, where a main licensee can invite other licensees to utilize the same license. Or for company-wide licenses for particular named-users.

ezekg:
Closes #534.

## Prerequisites

- [ ] Add API endpoints for assigning/unassigning users to licenses. (Should we use attach/detach?)
- [ ] Update `user` validation scope to look at `License#users`.
- [ ] Add `owner` validation scope.
- [x] Rename `License#user` to `License#owner`.
- [x] Rename singular `user` relationship to `owner` in user serializer.
- [x] Add plural `users` relationship to license serializer.
- [x] Add migration to rename `data.relationships.owner` to `data.relationship.user` for licenses (with updated links).
- [x] Bump API version to v1.6.
- [ ] ~~Rename `license.user.updated` webhook event to `license.owner.updated`? (Or send both?)~~
- [ ] Write integration tests for license owner vs license user authz.

## Pre-deploy

- [ ] Create `license.owner.updated`, `license.users.attached`, `license.users.detached`, `machine.owner.updated` event types.
- [ ] Create `license.owner.update`, `license.users.attach`, `license.users.detach`, `machine.owner.update` permissions.

## Post-deploy

- [ ] Rename `licenses.user_id` to `licenses.owner_id`? (May want to do this during our next maintenance window.)
- [ ] ~~Rename `license.user.update` permission to `license.owner.update`?~~
- [ ] Reseed database with new permissions/events.
- [ ] Add `license.owner.update`, `license.users.attach`, `license.users.detach`, `machine.owner.update` permissions to relevant roles.


app/models/license.rb
```diff
@@ -501,6 +508,28 @@ def permissions
...
   def entitlement_codes = entitlements.reorder(nil).codes
   def entitlement_ids   = entitlements.reorder(nil).ids
+  def user_ids          = users.reorder(nil).ids
+
+  # NOTE(ezekg) For backwards compatibility. Dual writing until we migrate
+  #             data from licenses.user_id column to HABTM.
+  def user=(record)
+    case record
+    in String => user_id
+      self.users = User.where(id: user_id)
```
ezekg, line 518: Needs to be account scoped.

app/models/license.rb
```diff
@@ -501,6 +508,28 @@ def permissions
...
+      self.users = [user]
+    in nil
+      self.users = []
+    end
+
+    super
+  end
+
+  def user_id=(id)
+    self.users = User.where(id:)
```
ezekg, line 529: Needs to be account scoped.

app/models/license.rb
ezekg: For time being, we could duplicate the `account_id`  attribute during license creation to make sure it's assigned first.

```rb
current_account.licenses.create(
  account_id: current_account.id,   
  **license_params,
)
```

db/scripts/seed_license_users.rb
```diff
@@ -0,0 +1,18 @@
...
+licenses = License.where_assoc_not_exists(:license_users)
+                  .where_assoc_exists(:user)
+                  .includes(:user)
+
+Rails.logger.info "Seeding #{licenses.count} license_users"
+
+licenses.find_each do |license|
+  Rails.logger.info "Seeding license_user for license=#{license.id} <> user=#{license.user_id}"
+
+  license.users << license.user
```
ezekg, line 15: Should probably add a sleep in here since this is going to be a large migration.

app/models/concerns/accountable.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+module Accountable
```
ezekg, line 3: We need to migrate all models to use `has_account` instead of `belongs_to :account`.

The main benefits here are better validations, and account inheritance via `default:`.

db/seeds/development.rb
```diff
@@ -70,6 +70,8 @@
         package = if rand(0..1).zero?
                     engine = if rand(0..1).zero?
                                { engine_attributes: { key: %w[pypi tauri].sample } }
+                             else
+                               {}
```
ezekg, line 74: This was raising an error when doing `ReleasePackage.create!(**engine)`...

app/models/machine.rb
```diff
@@ -614,23 +615,6 @@ def update_machines_core_count_on_destroy
     Keygen.logger.exception e
   end
 
-  def validate_associated_records_for_components
-    return if
-      components.nil? || components.empty?
-
-    components.each_with_index do |component, i|
-      component.account = account if
```
ezekg, line 622: The `Accountable` concern makes this hack unnecessary.

app/models/release.rb
```diff
@@ -808,23 +804,6 @@ def validate_associated_records_for_channel
     self.channel = rows.first
   end
 
-  def validate_associated_records_for_constraints
-    return if
-      constraints.nil? || constraints.empty?
-
-    constraints.each_with_index do |constraint, i|
-      constraint.account = account if
```
ezekg, line 816: This hack is longer needed due to `Accountable`'s default values.

app/models/user.rb
```diff
@@ -75,6 +85,35 @@ class User < ApplicationRecord
...
+  #     Semverse::Constraint.new('<= 1.5')
+  #                         .satisfies?(
+  #                           Semverse::Version.new(Current.api_version),
+  #                         )
+
+  #   destroy_association_async_job.perform_later(
+  #     owner_model_name: self.class.to_s,
+  #     owner_id: id,
+  #     association_class: _licenses.klass.to_s,
+  #     association_ids: _licenses.ids,
```
ezekg, line 111: This should only destroy the set of licenses ONLY if the set contains a single license.

app/models/concerns/accountable.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        before_validation unless: -> { account_id_attribute_assigned? || account_attribute_assigned? },
+          if: -> { new_record? && account_id.nil? },
+          on: %i[create],
+          &fn
+      end
+
+      # We also want to assert that the model's current account matches
+      # all of its :belongs_to associations that are accountable.
+      unless (reflections = reflect_on_all_associations(:belongs_to)).empty?
```
ezekg, line 79: Would be nice if we could warn on `belongs_to` associations defined after `has_account` is called.

(Ditto for the `Environmental` concern.)

spec/models/machine_spec.rb
```diff
@@ -54,7 +54,7 @@
   describe '#components_attributes=' do
     it 'should not raise when component is valid' do
       machine = build(:machine, account:, components_attributes: [
-        attributes_for(:component, account:),
+        attributes_for(:component, account:, environment: nil),
```
ezekg, line 57: Why doesn't `attributes_for` run through the factory's `initialize_with`? According to the docs, [it should](https://github.com/thoughtbot/factory_bot/blob/bd0a10c5ab9f93425beeba2ab17e503dceb368ab/docs/src/ref/build-strategies.md?plain=1#L55C1-L59).

ezekg: The problem here is that the `NIL_ACCOUNT` and `NIL_ENVIRONMENT` sentinel values aren't filtered out of the resulting attribute hash, which ofc causes issues during record initialization. Maybe we can patch `attributes_for` to reject those values like we are in `initialize_with`?

ezekg: Meh. Not worth the effort.

ezekg: Not worth the effort of a workaround, but I opened an issue: https://github.com/thoughtbot/factory_bot/issues/1606

app/models/license.rb
```diff
@@ -14,9 +15,12 @@ class License < ApplicationRecord
   belongs_to :policy
   belongs_to :group,
     optional: true
+  belongs_to :user,
+    optional: true
   has_one :product, through: :policy
   has_many :license_users, validate: false, index_errors: true, dependent: :destroy_async
-  has_many :users, index_errors: true, through: :license_users
+  has_many :licensees, index_errors: true, through: :license_users, source: :user
+  union_of :users, sources: %i[licensees user], class_name: User.name
```
ezekg, line 23: Using `class_name:` like this could result in load errors if dependencies are cyclic. Use a string instead.

app/models/license.rb
```diff
@@ -14,9 +15,12 @@ class License < ApplicationRecord
   belongs_to :policy
   belongs_to :group,
     optional: true
+  belongs_to :user,
```
ezekg, line 18: Rename to `owner`?

ezekg:
New plan: with the proof of concept `union_of` gem, we're able to create associations that union across multiple associations. This allows us to really clean up our implementation here, and gets rid of the need to do multi-part PRs.

Let's clean up `union_of` and add more exhaustive tests, and more in the direction mentioned in https://github.com/keygen-sh/keygen-api/issues/534#issuecomment-1858574833.

ezekg:
Should a machine belong to an owner? Right now, it doesn't make sense that a machine always belongs to the owner of the license, since in reality, it could belong to any of the license's licensees, not necessarily the owner.

PR #774, opened 2024-01-12
Add multi-user licenses

### LINKED ISSUE #534: A license should have multiple users

ezekg:
Renamed #768. Closes #534. TL;DR: licenses had-one user, now licenses have-many users.

## Prerequisites

- [x] Add API endpoints for assigning/unassigning users to licenses. (Should we use attach/detach?)
- [x] Update `user` validation scope to look at `License#users`.
- [ ] ~~Add `owner` validation scope? (May be too ambiguous with license vs machine owner.)~~
- [x] Rename `License#user` to `License#owner`.
- [x] Rename singular `user` relationship to `owner` in user serializer.
- [x] Add plural `users` relationship to license serializer.
- [x] Add migration to rename `data.relationships.owner` to `data.relationship.user` for licenses (with updated links).
- [x] Bump API version to v1.6.
- [ ] ~~Rename `license.user.updated` webhook event to `license.owner.updated`?~~ Not needed with v1x5 module.
- [x] Write integration tests for license owner vs license user authz.
- [ ] ~~Set default API version to v1.5 until post-deploy checklist is complete.~~ Workaround in 7a61cb0aab677b33effb0da6378b4f116b92cde2.
- [x] Get a third-party audit and review of `union_of`.
- [x] Write documentation for…
  - [x] New API endpoints and parameters
  - [x] New permissions
  - [x] New events
- [ ] Update dashboard app? (Could launch this with Portal instead of doing the extra work…)
- [x] Update changelog.

## Pre-deploy

- [x] Create `license.users.attached`, `license.users.detached`, `license.owner.updated`, `machine.owner.updated` event types.
- [x] Create `license.users.attach`, `license.users.detach`, `license.owner.update`, `machine.owner.update` permissions.

## Post-deploy

- [ ] ~~Rename `licenses.user_id` to `licenses.owner_id`? (May want to do this during our next maintenance window?)~~
- [ ] ~~Rename `license.user.update` permission to `license.owner.update`?~~ Not needed with v1x5 module.
- [ ] Run the below Rake tasks.
- [ ] ~~Set default API version to v1.6.~~ Workaround in 7a61cb0aab677b33effb0da6378b4f116b92cde2.
- [ ] Extract `union_of` to a gem.

```bash
# Add new permissions and event types
rake db:seed

# Migrate existing role permissions
rake keygen:permissions:admins:add[license.users.attach,license.users.detach,license.owner.update,machine.owner.update]
rake keygen:permissions:environments:add[license.users.attach,license.users.detach,license.owner.update,machine.owner.update]
rake keygen:permissions:products:add[license.users.attach,license.users.detach,license.owner.update,machine.owner.update]
```


lib/union_of.rb
```diff
@@ -0,0 +1,321 @@
...
+      INSTANCE.scope(association)
+    end
+
+    private
+
+    def last_chain_scope(scope, reflection, owner)
+      return super unless
+        reflection.union_of?
+
+      foreign_table = reflection.aliased_table
```
ezekg, line 41: Each union should have an alias derived from the association name, with a `union` suffix.

lib/union_of.rb
```diff
@@ -0,0 +1,321 @@
...
+    def join_scope(table, foreign_table, foreign_klass)
+      predicate_builder = predicate_builder(table)
+      scope_chain_items = join_scopes(table, predicate_builder)
+      klass_scope       = klass_join_scope(table, predicate_builder)
+
+      klass_scope.where!(
+        join_foreign_key => union_sources.reduce(nil) do |chain, union_source|
+          reflection = foreign_klass.reflect_on_association(union_source)
+          relation   = reflection.klass.scope_for_association.except(:order)
+                                                             .select(:id)
```
ezekg, line 182: Primary key may not always be `:id`.

lib/union_of.rb
```diff
@@ -0,0 +1,321 @@
...
+        ThroughAssociation
+      else
+        Association
+      end
+    end
+
+    def association_scope
+      return if klass.nil?
+
+      Scope.create(self)
```
ezekg, line 170: Should we cache this like Rails does?

lib/union_of.rb
```diff
@@ -0,0 +1,321 @@
...
+    def delete_all(...)  = raise NotImplementedError
+
+    def insert_record(...)  = raise NotImplementedError
+    def delete_records(...) = raise NotImplementedError
+    def concat_records(...) = raise NotImplementedError
+    def build_record(...)   = raise NotImplementedError
+  end
+
+  class Association < ReadonlyAssociation
+    def association_scope = Scope.scope(self)
```
ezekg, line 21: Should this check for `klass` and should it be cached?

lib/union_of.rb
ezekg: Actually, is this even needed?

app/controllers/api/v1/machine_components_controller.rb
```diff
@@ -13,7 +13,7 @@ class MachineComponentsController < Api::V1::BaseController
     before_action :set_machine_component, only: %i[show update destroy]
 
     def index
-      machine_components = apply_pagination(authorized_scope(apply_scopes(current_account.machine_components)).preload(:machine, :license, :policy, :product, :group, :user))
+      machine_components = apply_pagination(authorized_scope(apply_scopes(current_account.machine_components)).preload(:machine, :license, :policy, :product, :group, :owner))
```
ezekg, line 16: Preload `license: :owner`?

app/controllers/api/v1/machine_processes_controller.rb
```diff
@@ -14,7 +14,7 @@ class MachineProcessesController < Api::V1::BaseController
     before_action :set_machine_process, only: %i[show update destroy]
 
     def index
-      machine_processes = apply_pagination(authorized_scope(apply_scopes(current_account.machine_processes)).preload(:machine, :license, :policy, :product, :group, :user))
+      machine_processes = apply_pagination(authorized_scope(apply_scopes(current_account.machine_processes)).preload(:machine, :license, :policy, :product, :group, :owner))
```
ezekg, line 17: Preload `license: :owner`?

app/models/license.rb
```diff
@@ -2,21 +2,27 @@
...
   belongs_to :group,
     optional: true
+  belongs_to :owner,
+    foreign_key: :user_id,
+    class_name: User.name,
+    optional: true
   has_one :product, through: :policy
+  has_many :license_users, dependent: :destroy_async
+  has_many :licensees, through: :license_users, source: :user
+  union_of :users, sources: %i[licensees owner], class_name: User.name
```
ezekg, line 25: The `class_name` shouldn't be required.

app/models/license.rb
```diff
@@ -169,6 +176,19 @@ class License < ApplicationRecord
     errors.add :group, :license_limit_exceeded, message: "license count has exceeded maximum allowed by current group (#{group.max_licenses})"
   end
 
+  # Assert owner is not a license user (i.e. licensee)
+  validate on: %i[create update] do
+    next unless
+      user_id_changed?
```
ezekg, line 182: Create an alias for this: `owner_id_changed?`. Will make renaming the column in easier later on.

app/models/license.rb
```diff
@@ -408,7 +455,7 @@ class License < ApplicationRecord
     end
   }
   scope :banned, -> {
-    joins(:user).where.not(user: { banned_at: nil })
+    joins(:owner).where.not(users: { banned_at: nil })
```
ezekg, line 458: Should this be `owner:` to avoid ambiguity if also joined on `users`?

app/models/machine_component.rb
```diff
@@ -17,10 +17,13 @@ class MachineComponent < ApplicationRecord
     through: :license
   has_one :policy,
     through: :license
-  has_one :user,
+  has_one :owner,
```
ezekg, line 20: Should be through machine.

app/models/permission.rb
```diff
@@ -92,14 +93,17 @@ class Permission < ApplicationRecord
...
     license.user.update
+    license.users.attach
+    license.users.detach
     license.validate
 
     machine.check-out
     machine.create
     machine.delete
-    machine.group.update
+    machine.owner.update
```
ezekg, line 103: Sorting seems off here

app/models/product.rb
```diff
@@ -1,7 +1,9 @@
 # frozen_string_literal: true
 
 class Product < ApplicationRecord
+  include UnionOf::Macro
```
ezekg, line 4: ✂️

app/models/release.rb
```diff
@@ -281,15 +277,19 @@ class Release < ApplicationRecord
...
 
-    # Should we be applying a LIMIT to these UNION'd queries?
-    entl.joins(:users, :product)
+    entl.joins(product: %i[licenses])
       .where(
-        product: { distribution_strategy: ['LICENSED', nil] },
-        users: { id: user },
+        product: {
+          distribution_strategy: ['LICENSED', nil],
+          licenses: { id: License.for_user(user) },
```
ezekg, line 284: Is this needed? We can probably join on `users` instead since it's unioned.

app/models/user.rb
```diff
@@ -310,6 +325,7 @@ def entitlement_codes = entitlements.reorder(nil).codes
   def entitlement_ids   = entitlements.reorder(nil).ids
   def product_ids       = products.reorder(nil).ids
   def policy_ids        = policies.reorder(nil).ids
+  def license_ids       = licenses.reorder(nil).ids
```
ezekg, line 328: This isn't needee

app/policies/license_file_policy.rb
```diff
@@ -12,7 +12,7 @@ def show?
       allow!
     in role: Role(:product) if record.product == bearer
       allow!
-    in role: Role(:user) if record.user == bearer
+    in role: Role(:user) if record.owner == bearer
```
ezekg, line 15: Should include licensees.

lib/union_of.rb
ezekg: Meh. I'm not sure it's worth the effort.

app/models/license.rb
ezekg: Still have some ambiguity in the search methods, but not sure how I can work around that since Rails only aliases the join when your where clause includes a hash value, i.e. you can't do the following:

```rb
joins(:owner).where('owner.email ILIKE ?', '%example.com')
```

Instead, you have to do something like this:

```rb
joins(:owner).where.not(owner: { email: nil })
             .where('owner.email ILIKE ?', '%example.com')
```

I feel like you should be able to use `#references(:owner)` for this, but evidently that's not how it works.

spec/factories/plan.rb
```diff
@@ -4,7 +4,7 @@
   factory :plan do
     initialize_with { new(**attributes) }
 
-    name         { Faker::Company.buzzword }
+    name         { Keygen.ee? ? 'Ent 1' : 'Std 1' }
```
ezekg, line 7: This will give us coverage for both plan tiers i.e. Ent vs Std.

ezekg:
When we roll this out, we should set the `DEFAULT_API_VERSION` to v1.5 so we have ample time to run the permission tasks. Once things are ready and documentation has been updated, we can start putting new accounts into v1.6.

ezekg:
CI is failing here and on master because a Ruby 3.2.3 image isn't available yet.

app/controllers/api/v1/licenses/relationships/owners_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+      license.update!(user_id: owner_params[:id])
+
+      BroadcastEventService.call(
+        event: 'license.owner.updated',
+        account: current_account,
+        resource: license,
+      )
+
+      # FIXME(ezekg) This should be the user
```
ezekg, line 41: Technically, this should be the relationship linkage, but since this isn't under the `/relationships` namespace, it's a gray area. I still think that returning the license object here (and in other relationships) feels weird. Maybe we should update this to be the `owner` for both here and the machine's owner controller? Our users controller already returns the attached users for `#attach` so might as well take advantage of the fact we can fix our wrong here i.r.t. deprecating the old user controllers and replacing with these owner controllers.

ezekg: If we do go this route, changing the response, we'll need to assert the bearer has `user.read` permissions (which would be assumed since they also have `license.owner.update` permissions.

What would be nice here would be a `render jsonapi: ..., linkage: true` option, where only linkage is rendered.

ezekg: I'm going to disregard this for now. It's too much to diverge from what our other relationship endpoints return. Even if it's not perfect, it's what we have. If it's that much of a problem, we can always write a migration to fix it later.

app/models/machine_process.rb
```diff
@@ -90,10 +92,11 @@ class ResurrectionExpiredError < StandardError; end
...
 
-  scope :for_product, -> product { joins(license: :product).where(products: product) }
-  scope :for_license, -> license { joins(:license).where(licenses: license) }
-  scope :for_machine, -> machine { where(machine: machine) }
-  scope :for_user,    -> id { joins(:license).where(licenses: { user_id: id }) }
+  scope :for_product, -> id      { joins(license: :product).where(products: { id: }) }
+  scope :for_license, -> id      { joins(:license).where(licenses: { id: }) }
+  scope :for_machine, -> id      { joins(:machine).where(machine: { id: }) }
+  scope :for_user,    -> id      { joins(:users).where(users: { id: }) }
+  scope :for_owner,   -> id      { joins(:owner).where(owner: { id: }) }
```
ezekg, line 99: Fix formatting.

app/models/machine_component.rb
```diff
@@ -67,10 +70,11 @@ class MachineComponent < ApplicationRecord
...
 
-  scope :for_product, -> id { joins(:product).where(product: { id: }) }
-  scope :for_license, -> id { joins(:license).where(license: { id: }) }
-  scope :for_machine, -> id { joins(:machine).where(machine: { id: }) }
-  scope :for_user,    -> id { joins(:user).where(user: { id: }) }
+  scope :for_product, -> id      { joins(:product).where(product: { id: }) }
+  scope :for_license, -> id      { joins(:license).where(license: { id: }) }
+  scope :for_machine, -> id      { joins(:machine).where(machine: { id: }) }
+  scope :for_user,    -> id      { joins(:users).where(users: { id: }) }
+  scope :for_owner,   -> id      { joins(:owner).where(owner: { id: }) }
```
ezekg, line 77: Fix formatting.

app/policies/licenses/user_policy.rb
```diff
@@ -24,15 +44,40 @@ def show?
...
+        strict: license.environment.nil?,
+      )
+
+      case bearer
+      in role: Role(:admin | :developer | :sales_agent | :support_agent | :read_only | :environment)
+        allow!
+      in role: Role(:product) if license.product == bearer
+        allow!
+      in role: Role(:user) if license.owner == bearer
+        allow!
```
ezekg, line 61: We should only allow this when the policy or license is unprotected. This is a privilege escalation vector, where a license owner knows the ID of another user, but cannot read anything about them. Yet, if they attach that user to one of their licenses, they can now read the user because the user is now associated to them. Guessing IDs is out of the question, so this isn't a huge vector, but I still want this behavior to be explicit via an unprotected policy or license.

ezekg: Better yet, we could default user permissions to not allow attaching/detaching users to licenses. I prefer this approach, as it makes it even more explicit. We should still assert that the license is unprotected, in addition, since we do that everywhere else the license is mutated.

app/models/user.rb
```diff
@@ -306,11 +327,14 @@ class User < ApplicationRecord
     end
   }
 
+  # FIXME(ezekg) Selecting on ID isn't supported by our association scopes because
+  #              we're using DISTINCT and reordering on created_at.
+  def teammate_ids = teammates.reorder(nil).ids
```
ezekg, line 332: I feel like Rails should do this by default. Maybe look at a PR?

app/controllers/concerns/default_url_options.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+module DefaultUrlOptions
+  extend ActiveSupport::Concern
+
+  class_methods do
+    # FIXME(ezekg) Why is this needed?
+    def default_url_options(...) = Rails.application.default_url_options(...)
```
ezekg, line 8: Why is this needed? And why was the previous way of doing this failing only when the `CI` environment variable is set?!

app/services/license_checkout_service.rb
```diff
@@ -11,6 +11,7 @@ class InvalidLicenseError < StandardError; end
     policy
     group
     owner
+    users
```
ezekg, line 14: It would be nice to add a limit to how many users (and entitlements) can be included.

lib/union_of.rb
```diff
@@ -0,0 +1,299 @@
...
+
+  module Macro
+    extend ActiveSupport::Concern
+
+    class_methods do
+      def union_of(name, scope = nil, **options, &extension)
+        reflection = Builder.build(self, name, scope, options, &extension)
+
+        ActiveRecord::Reflection.add_union_reflection(self, name, reflection)
+        ActiveRecord::Reflection.add_reflection(self, name, reflection)
```
ezekg, line 225: I want to hire somebody to review and audit this gem.

lib/union_of.rb
```diff
@@ -0,0 +1,299 @@
+# frozen_string_literal: true
+
+require 'active_record_union'
+
+module UnionOf
```
ezekg, line 5: I want to hire somebody to review and audit this gem before we merge.

lib/union_of.rb
```diff
@@ -0,0 +1,299 @@
+# frozen_string_literal: true
+
+require 'active_record_union'
```
ezekg, line 3: Is this needed?

lib/union_of.rb
```diff
@@ -0,0 +1,299 @@
+# frozen_string_literal: true
```
ezekg, line 1: Is this dependency needed? IIRC, we're using Arel everywhere we were previously using this gem.

lib/union_of.rb
ezekg, line 3: Is this dependency needed? IIRC, we're using Arel everywhere we were previously using this gem.

lib/union_of.rb
ezekg: Shortlist:

- https://andyatkinson.com/consulting
- https://spektrhq.com/consultancy

db/seeds/development.rb
```diff
@@ -124,26 +126,41 @@
...
-                     environment:,
-                     account:,
-                   )
-                 end
-
-          if rand(0..1).zero? && user.present?
-            Token.create!(bearer: user, account:, environment:)
-          end
-
           license = License.create!(
```
ezekg, line 129: Need to randomly add an owner to seeded licenses.

spec/shared/accountable.rb
```diff
@@ -0,0 +1,72 @@
...
+        after  { Current.account = nil }
+
+        it 'should raise when account is nil' do
+          expect { create(factory, account: nil) }.to raise_error ActiveRecord::RecordInvalid
+        end
+
+        it 'should set provided account' do
+          instance = create(factory, account:)
+
+          expect(instance.account).to eq account
```
ezekg, line 41: This test is useless because `Current.account == account`. To resolve, let's set `Current.account` to a different account.

ezekg (commented):
First review looks good. Still need to get an audit for `union_of`.

app/models/license.rb
```diff
@@ -367,6 +415,13 @@ class License < ApplicationRecord
       where 'user_id IS NOT NULL'
     end
   }
+  scope :assigned, -> (status = true) {
+    if ActiveRecord::Type::Boolean.new.cast(status)
+      where 'user_id IS NOT NULL'
```
ezekg, line 420: Should this check `where.associated(:users)` instead?

ezekg:
Currently waiting on audit and review of `union_of`.

lib/union_of.rb
ezekg, line 5: Currently, eager loading is not supported. Should we add support or raise?

ezekg: Here's where we can do an extension: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/associations/preloader/branch.rb#L137

lib/union_of.rb
```diff
@@ -0,0 +1,422 @@
...
+                         )
+                         .where(
+                           through_table[union_source_foreign_key].in(keys),
+                         )
+                  when union_source_reflection.belongs_to?
+                    table.project(table[primary_key])
+                         .from(table)
+                         .join(foreign_table, Arel::Nodes::InnerJoin)
+                         .on(
+                           table[union_source_foreign_key].eq(foreign_table[primary_key]),
```
ezekg, line 95: This is using the wrong primary key. It should be `foreign_reflection.active_record_primary_key`.

lib/union_of.rb
ezekg: Correction: eager loading does not work because there's no way to resolve the inverses of the eager loaded records without loading the associations themselves.

spec/support/matchers/sql_matcher.rb
```diff
@@ -0,0 +1,27 @@
...
+
+RSpec::Matchers.define :match_sql do |expected|
+  # NOTE(ezekg) Make our expected SQL string match #to_sql format.
+  #             This mostly deals with formatting parentheses.
+  expected = expected.dup.tap do |s|
+    s.squish!
+    s.gsub!(/(\()\s*([\w'"])/, '\1\2')     # `( "table"."column"` => `("table"."column"`
+    s.gsub!(/([\w"'])\s*(\))/, '\1\2')     # `"table"."column" = 'value')` => `"table"."column" = 'value')`
+    s.gsub!(/\s+(\()\s+(\()\s+/, ' \1\2 ') # ` ( ( ` => ` (( `
+    s.gsub!(/\s+(\))\s+(\))\s+/, ' \1\2 ') # ` ) ) ` => ` )) `
```
ezekg, line 11: Use https://github.com/alekseyl/niceql instead.

lib/union_of.rb
ezekg: Eager loading needs to be solved due to #792.

spec/lib/union_of_spec.rb
```diff
@@ -343,6 +334,246 @@
...
+          "machines"."license_id" AS t1_r10,
+          "machines"."last_heartbeat_at" AS t1_r11,
+          "machines"."cores" AS t1_r12,
+          "machines"."last_death_event_sent_at" AS t1_r13,
+          "machines"."group_id" AS t1_r14,
+          "machines"."max_processes_override" AS t1_r15,
+          "machines"."last_check_out_at" AS t1_r16,
+          "machines"."environment_id" AS t1_r17,
+          "machines"."heartbeat_jid" AS t1_r18,
+          "machines"."owner_id" AS t1_r19
```
ezekg, line 512: This is brittle and will break if we add a column to these tables.

spec/lib/union_of_spec.rb
```diff
@@ -343,6 +334,246 @@
...
+          "users"."password_reset_sent_at" AS t1_r6,
+          "users"."metadata" AS t1_r7,
+          "users"."account_id" AS t1_r8,
+          "users"."first_name" AS t1_r9,
+          "users"."last_name" AS t1_r10,
+          "users"."stdout_unsubscribed_at" AS t1_r11,
+          "users"."stdout_last_sent_at" AS t1_r12,
+          "users"."banned_at" AS t1_r13,
+          "users"."group_id" AS t1_r14,
+          "users"."environment_id" AS t1_r15
```
ezekg, line 436: This is brittle and will break if we add a column to these tables.

lib/union_of.rb
```diff
@@ -0,0 +1,299 @@
...
+
+module UnionOf
+  class ReadonlyAssociationError < ActiveRecord::ActiveRecordError
+    def initialize(owner, reflection)
+      super("Cannot modify association '#{owner.class.name}##{reflection.name}' because it is read-only.")
+    end
+  end
+
+  class ReadonlyAssociation < ActiveRecord::Associations::CollectionAssociation
+    def writer(...)         = raise ReadonlyAssociationError.new(owner, reflection)
```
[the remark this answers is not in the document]
gregmolnar: You could use define_method with an array of methods since they all do the same.

lib/union_of.rb
```diff
@@ -0,0 +1,411 @@
...
+module UnionOf
+  class Error < ActiveRecord::ActiveRecordError; end
+
+  class ReadonlyAssociationError < Error
+    def initialize(owner, reflection)
+      super("Cannot modify association '#{owner.class.name}##{reflection.name}' because it is read-only")
+    end
+  end
+
+  class ReadonlyAssociation < ActiveRecord::Associations::CollectionAssociation
```
ezekg, line 12: Need to add tests for this.

lib/union_of.rb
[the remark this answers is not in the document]
ezekg: Implemented in https://github.com/keygen-sh/keygen-api/pull/774/commits/3e8f4076c42ef3fffc44ca8850925bdb11cda8d2.

app/services/license_validation_service.rb
```diff
@@ -75,9 +75,15 @@ def validate!
 
       # Check against :user scope requirements
       if scope.present? && scope.key?(:user)
+        user_identifier = scope[:user]
+
         return [false, "user scope does not match", :USER_SCOPE_MISMATCH] unless
-          license.user&.email == scope[:user] ||
-          license.user&.id == scope[:user]
+          license.owner_id == user_identifier ||
```
ezekg, line 81: Should providing a user scope also scope any machine fingerprints to that user via `machine.owner`?

Basically, do we want user A to be able to validate machine X when machine X belongs to user B, not user A?

app/policies/licenses/v1x5/user_policy.rb
```diff
@@ -0,0 +1,41 @@
...
+        allow!
+      in role: Role(:license) if license == bearer
+        allow!
+      else
+        deny!
+      end
+    end
+
+    def update?
+      verify_permissions!('license.user.update')
```
ezekg, line 28: Just a note to future self: this is backwards compatible with the new `license.owner.update` permission, since requests to `PUT /licenses/<license>/user` use this policy and not the new owner policy.

PR #791, opened 2024-02-07
Fix N+1 on users index when users are not new (Prosopite)

ezekg:
Resolves an N+1 issue caused by `User#status` for the users index when the collection contained old users. Also replaces Bullet with Prosopite because Bullet missed this, and I couldn't force the test failure, and it failed right away with Prosopite.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -350,16 +350,16 @@ class User < ApplicationRecord
   end
 
   def entitled?(*identifiers)
-    identifiers = identifiers.flatten
-                             .compact
-
+    entls = entls.flatten.compact
     return true if
-      identifiers.empty?
+      entls.empty?
 
-    entls = Entitlement.where(id: identifiers)
-                       .or(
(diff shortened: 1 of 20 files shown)


app/models/user.rb
```diff
@@ -30,6 +30,16 @@ class User < ApplicationRecord
...
     through: :group_owners
 
+  # NOTE(ezekg) This association is only used to preload a user's status, since
+  #             the #status needs to check if a user has any active licenses.
+  #
+  #             We're doing a DISTINCT ON so that we can query ANY one active
+  #             license for each user, since users can potentially have thousands
+  #             and thousands of licenses, and superfluously preloading/querying
+  #             that many records would a bad idea.
+  has_one :any_active_license, -> { active.reorder(nil).distinct_on(:user_id) },
```
ezekg, line 40: I love Postgres. This would've been so complicated with MySQL using a `LATERAL JOIN`.

ezekg:
Closing in favor of #792.

PR #799, opened 2024-02-24
Add revision header and logging

ezekg:
- [x] Add a `Keygen-Revision` header for CE/EE audit purposes.
- [x] Add to boot logging,
- [x] Add to lograge.


ezekg:
Add to lograge.

lib/keygen/version.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+
+      private
+
+      def from_environment = ENV['COMMIT_HASH'] || ENV['COMMIT_SHA'] || ENV['HEROKU_SLUG_COMMIT']
+      def from_git         = `git rev-parse HEAD` rescue nil
+      def from_file(path = Rails.root / 'VERSION')
+        return unless File.exist?(path)
+
+        File.read(path).chomp
```
ezekg, line 30: This should filter out `$Id$` in case it wasn't replaced with `ident` by git.

ezekg: This is the resulting `VERSION` file whenever the repo is cloned:

```
$Id: 055c8729cdcc372500a08db659c045e16c4409fb $
```

So a regexp would work best here.

VERSION
```diff
@@ -0,0 +1 @@
+$Id$
```
ezekg, line 1: I'm not the biggest fan of this. I feel it's going to be a chore to not accidentally commit changes to this file.

Is this even needed?

.gitattributes
```diff
@@ -0,0 +1 @@
+/VERSION ident
```
ezekg, line 1: https://git-scm.com/docs/gitattributes#_ident

VERSION
ezekg: Ah, apparently this is a moot point. You can't commit the `$Id: ... $` changeset. 😅

PR #808, opened 2024-04-07
fix: use postgresql-client in the production image

aminya:
The production image doesn't need the development files of Postgres. By replacing it with `postgresql-client`, the size of the container image is reduced significantly.

This reduces the size of the image by 512 MB

- After `206MB`
- Before `718MB`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -38,7 +38,7 @@ COPY . /app
 
 RUN apk add --no-cache \
   bash \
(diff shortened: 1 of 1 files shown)


ezekg:
Thanks for the PR! Failed build is related to secrets not being shared with forks.

aminya:
You're welcome!
Good to know.

PR #813, opened 2024-04-19
perf: optimizations for the Docker image

aminya:
- [fix: optimize the RUN commands order in Docker](https://github.com/keygen-sh/keygen-api/commit/77ec406989a928679ef6ccf9b0e9329634e81b57): reorders and merges the RUN commands to reduce layer sizes and also speed up the builds when a file changes during development
- [fix: use bundle config for setting bundle options](https://github.com/keygen-sh/keygen-api/commit/96775ddbf9913d0e608e4f74b091b789b7cd7b5f): uses `bundle config` for setting the bundle settings, which is the recommended approach as per bundle's documentation. It also allows the bundle to use all the CPU cores during the builds.
- [fix: remove unused C/C++ build remainders](https://github.com/keygen-sh/keygen-api/commit/3ca9ba99c9b1d7049328816ceeaae0e2ef6a8704): removes some C/C++ build remainders after bundle install
- ~[fix: expose the specified port in docker](https://github.com/keygen-sh/keygen-api/pull/813/commits/fb5932a42aba1bfe116aaf81dec04025562b96d3)~

These reduce the docker size image by 10 MB (5%).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -63,14 +63,11 @@ ENV KEYGEN_EDITION="CE" \
     PORT="3000" \
     BIND="0.0.0.0"
 
-RUN \
-  chmod +x /app/scripts/entrypoint.sh && \
+RUN chmod +x /app/scripts/entrypoint.sh && \
   adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen && \
   chown -R keygen:keygen /app
 
 USER keygen
-
-EXPOSE $PORT
 
 ENTRYPOINT ["/app/scripts/entrypoint.sh"]
 CMD ["web"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,17 +24,13 @@ RUN apk add --no-cache \
   tzdata \
   openssl \
   postgresql-dev \
-  libc6-compat
-
-RUN bundle install --jobs 4 --retry 5 && \
+  libc6-compat && \
+  bundle install --jobs 4 --retry 5 && \
   chmod -R a+r /usr/local/bundle
 
 # Final stage
 FROM base
 LABEL maintainer="keygen.sh <oss@keygen.sh>"
-
-WORKDIR /app
-COPY . /app
 
 RUN apk add --no-cache \
   bash \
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -24,38 +24,53 @@ RUN apk add --no-cache \
...
+  chmod +x /app/scripts/entrypoint.sh && \
   adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen && \
   chown -R keygen:keygen /app
 
-COPY --from=build --chown=keygen:keygen \
-  /usr/local/bundle/ /usr/local/bundle
-
 USER keygen
 
+EXPOSE $PORT
```
ezekg, line 73: The port is only relevant for the web process, which is why it wasn't here before. For example, the worker process should not be exposed on a port.

Dockerfile
```diff
@@ -24,38 +24,53 @@ RUN apk add --no-cache \
...
+
 ENV KEYGEN_EDITION="CE" \
     KEYGEN_MODE="singleplayer" \
     RAILS_LOG_TO_STDOUT="1" \
     PORT="3000" \
     BIND="0.0.0.0"
 
-RUN chmod +x /app/scripts/entrypoint.sh && \
+RUN \
+  chmod +x /app/scripts/entrypoint.sh && \
```
ezekg, line 67: Drop this change.

ezekg (changes requested):
Thanks. Just a couple small changes. 👍

Dockerfile
aminya: That makes sense. However, there's no downside to exposing it for other cases. It will be simply not used.

aminya: Reverted this any way to unblock this:
[`01ec10e` (#813)](https://github.com/keygen-sh/keygen-api/pull/813/commits/01ec10ed3bfbd7455117e27a5a0e54f6b168a158)

But we can discuss bringing it back.

Dockerfile
```diff
@@ -52,9 +67,6 @@ RUN chmod +x /app/scripts/entrypoint.sh && \
   adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen && \
   chown -R keygen:keygen /app
 
-COPY --from=build --chown=keygen:keygen \
```
ezekg, line 55: Does it matter that `COPY --chown=keygen:keygen` is run before `RUN adduser -g keygen -D keygen` here?

Why was the `COPY` moved higher up? Just for my own edification.

ezekg:
Ignore the build failures. Unfortunately, GitHub won't let me run the CI with secrets for approved PRs.

Dockerfile
aminya: The difference is that the `COPY .` command is right before those files are needed. This makes the rebuilds faster when a file changes in the repository as Docker caches all the previous steps including this bundle COPY.

Although I missed this chown here, which is now fixed in [`57dd1f6` (#813)](https://github.com/keygen-sh/keygen-api/pull/813/commits/57dd1f638ddfc79b6d56bac2d5463cd2bf91f07a)

Dockerfile
```diff
@@ -24,38 +24,50 @@ RUN apk add --no-cache \
...
-WORKDIR /app
-COPY . /app
-
 RUN apk add --no-cache \
   bash \
   postgresql-client \
   tzdata \
-  libc6-compat
+  libc6-compat && \
+  adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen
```
ezekg, line 53: Let's move this into its own `RUN` command so that dependencies and permissions are clearly separate.

aminya: Merging the RUN commands is the best practice recommended. I am using Hadolint, and it gives me a warning if I do that.  Should I proceed? 
https://github.com/hadolint/hadolint/wiki/DL3059#multiple-consecutive-run-instructions-consider-consolidation

ezekg: Ah, okay. No worries then. It was a nit anyways.

ezekg (approved):

PR #815, opened 2024-04-19
Fix nullable foreign keys (again)

ezekg:
I guess I gotta do everything twice these days. I shouldn't have been too hasty to deploy #814, since changing a column to `NOT NULL` in Postgres locks the entire table for _reads and writes_. This took down production for a few minutes until I could kill the locking queries and revert things. This new PR follows recommendations in [`strong_migrations`](https://github.com/ankane/strong_migrations?tab=readme-ov-file#setting-not-null-on-an-existing-column).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,6 +51,7 @@ gem 'kaminari', '~> 1.2.0'
 gem 'active_record_union'
 gem 'active_record_distinct_on', '~> 1.6'
 gem 'activerecord_where_assoc', '~> 1.1.4'
+gem 'strong_migrations'
 
 # Pattern matching
 gem 'rails-pattern_matching'
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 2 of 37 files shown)


Gemfile
```diff
@@ -51,6 +51,7 @@ gem 'kaminari', '~> 1.2.0'
 gem 'active_record_union'
 gem 'active_record_distinct_on', '~> 1.6'
 gem 'activerecord_where_assoc', '~> 1.1.4'
+gem 'strong_migrations'
```
ezekg, line 54: This will help prevent this from happening again.

db/schema.rb
```diff
@@ -656,7 +656,7 @@
   end
 
   create_table "request_logs", id: :uuid, default: -> { "uuid_generate_v4()" }, force: :cascade do |t|
-    t.uuid "account_id"
+    t.uuid "account_id", null: false
```
ezekg, line 659: This is out of sync. I didn't add this one to this PR. Likely from the previous PR. Will need to reset the schema.

ezekg:
Closing this in favor of doing each of these in separate PRs to de-risk.

Issue #818, opened 2024-04-20
Update nullable foreign keys to be non-nullable


ezekg:
For correctness, and for query performance via query planner optimizations.

- [ ] `billings.account_id`
- [ ] `keys.account_id`
- [ ] `keys.policy_id`
- [ ] `licenses.account_id`
- [ ] `licenses.policy_id`
- [ ] `licenses.product_id` (#817)
- [ ] `machines.account_id`
- [ ] `machines.license_id`
- [ ] `metrics.account_id`
- [ ] `policies.account_id`
- [ ] `policies.product_id`
- [ ] `products.account_id`
- [ ] `request_logs.account_id`
- [ ] `roles.resource_id`
- [ ] `roles.resource_type`
- [ ] `tokens.account_id`
- [ ] `tokens.bearer_id`
- [ ] `tokens.bearer_type`
- [ ] `users.account_id`
- [ ] `webhook_endpoints.account_id`
- [ ] `webhook_events.account_id`

PR #819, opened 2024-04-20
Update licenses.account_id to be NOT NULL

ezekg:
Part of #818. Extracted from #815. Updates config for `strong_migrations` to add default timeouts. Adds `verbose_migrations`.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -58,6 +58,10 @@ module Keygen
     # See above comment about having to use this multiple
     config.middleware.use Keygen::Middleware::RequestErrorWrapper
 
+    # Use the lowest log level to ensure availability of diagnostic information
+    # when problems arise.
+    config.log_level = ENV.fetch('RAILS_LOG_LEVEL') { :info }.to_sym
+
     # FIXME(ezekg) Should we migrate to credentials?
     config.active_record.encryption.primary_key         = ENV.fetch('ENCRYPTION_PRIMARY_KEY')
     config.active_record.encryption.deterministic_key   = ENV.fetch('ENCRYPTION_DETERMINISTIC_KEY')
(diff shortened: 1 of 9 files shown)


db/migrate/20240420053540_add_account_id_not_null_constraint_for_licenses.rb
```diff
@@ -0,0 +1,9 @@
+class AddAccountIdNotNullConstraintForLicenses < ActiveRecord::Migration[7.1]
+  def up
+    log_level_was, ActiveRecord::Base.logger.level = ActiveRecord::Base.logger.level, Logger::DEBUG
```
ezekg, line 3: Should we extract this out into an extension? Seems like we would want to use it for most migrations.

ezekg: Squatting https://rubygems.org/gems/verbose_migrations for this. Right now, thinking of a simple API like:

```rb
class AddAccountIdNotNullConstraintForLicenses < ActiveRecord::Migration[7.1]
  verbose! # or: enable_verbose_logging!

  def up
    add_check_constraint :licenses, 'account_id IS NOT NULL', name: 'licenses_account_id_not_null', validate: false
  end
end
```

lib/verbose_migrations.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'logger'
+
+module VerboseMigrations
+  module MigrationExtension
+    cattr_accessor :verbosity, default: nil
+
+    def verbose? = verbosity.present? && ActiveRecord::Base.logger.present?
+    def verbose!(level: Logger::DEBUG)
```
ezekg, line 10: Add `verbose_logger: ActiveRecord::Base.logger` so the logger is configurable, since we may want to e.g. make `Keygen.logger` verbose at times.

config/application.rb
```diff
@@ -58,6 +58,10 @@ class Application < Rails::Application
     # See above comment about having to use this multiple
     config.middleware.use Keygen::Middleware::RequestErrorWrapper
 
+    # Use the lowest log level to ensure availability of diagnostic information
+    # when problems arise.
+    config.log_level = ENV.fetch('RAILS_LOG_LEVEL') { :info }.to_sym
```
ezekg, line 63: Downcase

ezekg:
Happened to deploy during an autovacuum, so the lock failed. Redeploying was successful.

PR #838, opened 2024-05-09
Add PER_USER leasing strategy to policies

### LINKED ISSUE #831: Add a PER_USER leasing strategy to policies
To enforce process limits on a per-user basis. Should only be applied when the process's machine has an owner.

ezekg:
Closes #831.

### DIFF
--- a/app/models/machine.rb
+++ b/app/models/machine.rb
@@ -122,52 +122,102 @@ class Machine < ApplicationRecord
   validate on: :create do |machine|
     next if machine.license.nil?
     next if
-      machine.license.max_machines.nil? || machine.license.always_allow_overage?
+      machine.license.always_allow_overage?
 
-    prev_machines_count = machine.license.machines_count || 0
-    next if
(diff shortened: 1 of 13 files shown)


app/models/machine.rb
```diff
@@ -124,50 +124,94 @@ class ResurrectionExpiredError < StandardError; end
...
+
+      next if
+        machine.license.allow_1_5x_overage? && next_core_count <= machine.license.max_cores * 1.5
+
+      next if
+        machine.license.allow_2x_overage? && next_core_count <= machine.license.max_cores * 2
+
+      machine.errors.add :base, :core_limit_exceeded, message: "machine core count has exceeded maximum allowed for license (#{machine.license.max_cores})"
+    when license.lease_per_user?
+      prev_core_count = machine.license.machines.where.not(id: machine.id)
```
ezekg, line 197: Should this be scoped to product as well? This makes it so the same machine can't be used across multiple products for a given owner.

PR #839, opened 2024-05-09
Add PER_USER uniqueness strategies to policies

### LINKED ISSUE #837: Add UNIQUE_PER_USER fingerprint uniqueness strategies
For both machines and components.

ezekg:
Closes #837. Need to rebase after #838 is merged.

### DIFF
--- a/app/models/machine.rb
+++ b/app/models/machine.rb
@@ -14,7 +14,7 @@ class Machine < ApplicationRecord
   include Dirtyable
 
   HEARTBEAT_DRIFT = 30.seconds
-  HEARTBEAT_TTL = 10.minutes
+  HEARTBEAT_TTL   = 10.minutes
 
   belongs_to :license, counter_cache: true
   belongs_to :owner,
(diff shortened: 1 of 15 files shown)


app/models/machine.rb
```diff
@@ -223,6 +223,12 @@ class ResurrectionExpiredError < StandardError; end
       errors.add :fingerprint, :taken, message: "has already been taken for this product" if account.machines.joins(:product).exists?(fingerprint:, products: { id: product.id })
     when unique_per_policy?
       errors.add :fingerprint, :taken, message: "has already been taken for this policy" if account.machines.joins(:policy).exists?(fingerprint:, policies: { id: policy.id })
+    when unique_per_user?
+      errors.add :fingerprint, :taken, message: "has already been taken for this user" if
+        account.machines.exists?(
+          fingerprint:,
+          owner:,
```
ezekg, line 230: Feel like this needs to be scoped to the product so that machines across multiple products can work for the same owner.

ezekg:
I'm going to close this for now because after thinking on it more, I can't really think of a use-case for this that `PER_LICENSE` doesn't already accomplish. If somebody needs it and has a valid use case, I'll revisit. As is, this just adds complexity.

PR #843, opened 2024-05-17
Upgrade to Rails v7.2

ezekg:
WIP. Using the current release candidate. ~~Bullet doesn't support Rails v7.2 yet.~~


Gemfile
```diff
@@ -109,7 +109,7 @@ group :development, :test do
   gem 'byebug', platform: :mri
   gem 'dotenv-rails'
   gem 'timecop', '~> 0.9.5'
-  gem 'bullet', '~> 7.1.6'
+  # gem 'bullet', '~> 7.1.6'
```
ezekg, line 112: https://github.com/flyerhzm/bullet/issues/705

config/initializers/active_record_where_assoc.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module ActiveRecordWhereAssoc
+  module ActiveRecordCompat
+    if ActiveRecord.gem_version >= Gem::Version.new('7.2.0.alpha')
+      def self.normalize_association_name(name) = name.to_sym
```
ezekg, line 6: Make a PR for this.

ezekg: https://github.com/MaxLap/activerecord_where_assoc/issues/14

config/application.rb
```diff
@@ -80,6 +80,36 @@ class Application < Rails::Application
...
+    # I prefer to keep this explicit:
+    #
+    #   product = Product.new(
+    #     releases: [Release.new(product:)],
+    #   )
+    #
+    config.active_record.automatically_invert_plural_associations = false
+
+    # Implicitly define inverses for has-many associations
+    config.active_record.has_many_inversing = true
```
ezekg, line 108: Took 3 days to fully figure this out. See above comment for full explanation of the "unexpected" behavior. I understand it's "correct" i.r.t. inverses, but — to me at least — it doesn't make sense to save associated records implicitly. What if one of those records were discarded, e.g. a default that is later replaced by a new record during initialization? May seem like a contrived example, but that happens all the time in factory-based tests.

spec/factories/entitlement.rb
```diff
@@ -17,15 +17,15 @@
     name { code.humanize }
 
     trait :in_isolated_environment do
-      environment { build(:environment, :isolated, account:) }
+      environment { create(:environment, :isolated, account:) }
```
ezekg, line 20: Due to the way Rails v7.2 handles collection association inverses, the environment needs to be fully created before the other records are created, otherwise when the environment is saved, it also attempts to save any new associated records and that explodes because the environment is nil on those records.

E.g. given these records:

```ruby
environment = Environment.new
product     = Product.new(environment:)
```

The following returns a collection including the initialized product:

```ruby
environment.products # => [#<Product id=nil>]
```

In addition, when the environment is saved, so are all of its implicit associations:

```ruby
environment.save

environment.persisted? # => true
product.persisted?     # => true
```

Creating the environment first avoids this issue.

ezekg: For reference, look at `config.active_record.automatically_invert_plural_associations` and `config.active_record.has_many_inversing`, both of which default to `true` in Rails v7.2

spec/models/policy_spec.rb
```diff
@@ -75,6 +75,9 @@
     end
 
     context 'on update' do
+      before { Sidekiq::Testing.inline! }
+      after  { Sidekiq::Testing.fake! }
```
ezekg, line 79: Not sure what changes internally for Active Job to all of a sudden require this, but w/e.

spec/factories/entitlement.rb
ezekg: I think this may be a bug: https://github.com/rails/rails/issues/51853

spec/models/policy_spec.rb
zzak: @ezekg I wonder if this is related to rails/rails#48585

ezekg: @zzak that's it! Makes sense now that I think about it. Thanks.

app/models/concerns/environmental.rb
```diff
@@ -61,15 +61,18 @@ module Environmental
     #
     # Use :default to automatically configure a default environment for the model.
     # Accepts a proc that resolves into an Environment or environment ID.
-    def has_environment(default: nil)
-      belongs_to :environment,
-        optional: true
+    def has_environment(default: nil, **kwargs)
+      belongs_to :environment, optional: true, inverse_of: nil, **kwargs
```
ezekg, line 65: I thought setting `inverse_of: nil` would resolve our issue but doesn't look like it did. We should rip this out until the possible bug[^1] in Rails is resolved.

[^1]: https://github.com/keygen-sh/keygen-api/pull/843/files#r1605536192

app/models/concerns/environmental.rb
```diff
@@ -153,6 +156,10 @@ def has_environment(default: nil)
             next unless
               association.class < Environmental
 
+            # See: https://github.com/rails/rails/issues/51853
+            environment_id = read_attribute(:environment_id) ||
+                             environment&.id
```
ezekg, line 161: This is a hack for https://github.com/rails/rails/issues/51853

ezekg: Doesn't work either. I think I'm blocked by the change in order of operations right now. 😩

Gonna see if I can figure out how to fix upstream instead of trying workarounds.

ezekg:
I tried various workarounds for https://github.com/rails/rails/issues/51853, but I think this is a bug that needs to be fixed upstream.

zzak:
@ezekg Are you willing to try and bisect with your repro script to help identify the cause of the change?

ezekg:
> @ezekg Are you willing to try and bisect with your repro script to help identify the cause of the change?

Sure. I posted the bisect results in https://github.com/rails/rails/issues/51853#issuecomment-2122872428.

ezekg:
I think failing tests are related to the export/import test missing:

```rb
before { Sidekiq::Testing.inline! }
after  { Sidekiq::Testing.fake! }
```

ezekg:
Blocked by https://github.com/alecdotninja/active_record_distinct_on/issues/30.

PR #853, opened 2024-05-29
Add configurable statement_timeout to prune and vacuum workers

ezekg:
Some of these queries need a longer `statement_timeout` than our default. This makes the timeouts configurable.

### DIFF
--- a/lib/statement_timeout.rb
+++ b/lib/statement_timeout.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module StatementTimeout
+  module QueryMethodsExtension
+    def statement_timeout(timeout)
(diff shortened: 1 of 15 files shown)


lib/statement_timeout.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module StatementTimeout
+  module QueryMethodsExtension
+    def statement_timeout(timeout)
```
ezekg, line 5: Would be cool to eventually support usage without a block, too. E.g.:

```rb
License.statement_timeout(1.minute).some_heavy_report
```

I reserved the `statement_timeout` gem name, too.

PR #858, opened 2024-07-01
fix(api): update rate limiting docs URL

Pinjasaur:
Accidentally spammed the API a few too many times and realized the included `errors` URL isn't correct any more. Otherwise, helpful docs. 🙂


ezekg:
Appreciate it! I've been meaning to update this but I kept forgetting.

PR #863, opened 2024-07-15
Add SAML/SSO

### LINKED ISSUE #409: Add SAML/SSO
Let's use WorkOS for this. Seems [pretty simple to integrate](https://workos.com/docs/sso/2-add-sso-to-your-app). Pricing also [looks good](https://workos.com/pricing).

### Redirect

Use the normal create token endpoint. Add an `SSO_REQUIRED` error code when the current account or user requires SSO.

```
POST /v1/tokens
401 Unauthorized
Location: <authorization_url>
{
  "errors": [
    {
      "title": "Unauthorized",
      "detail": "single sign on is required",
      "code": "SSO_REQUIRED"
    }
  ],
  "links": {
    "redirect": "<authorization_url>"
  }
}
```

The client i.e. Portal MUST manually redirect to the WorkOS authorization URL.

Docs: https://workos.com/docs/reference/sso/get-authorization-url

### Callback

The callback is contains a `code` query param that can be exchanged for a user profile. After retrieving the user, we MAY need to create the user, and then we MUST generate a token for the authenticated user.

We can optionally use the `state` query param to store additional state, such as account or redirect URL.

After the user has been authenticated, we redirect to Portal.

```
GET /v1/workos?code=<code>&state=<account>
301 Temporary Redirect
Location: https://portal.keygen.sh/<account>
```

I'm thinking we should generate an SSO-specific token that can only be used via Portal e.g. using a strict/secure cookie. ~~But using cookies creates other problems, such as accounts using a self-hosted Portal not being able to use SSO. (We COULD keep the `Referrer` in the `state` parameter of the authorization URL, but that introduces an open redirect attack vector...)~~ If this becomes a problem, we can manually set a `sso_redirect_url` column for accounts using SSO.

Docs: https://workos.com/docs/reference/sso/get-authorization-url/redirect-uri

ezekg:
Closes #409. Implements SAML/SSO using [WorkOS](https://workos.com). Implements [JIT-provisioning](https://workos.com/docs/sso/jit-provisioning) for new users.

This includes support for the following:

- Google OAuth
- GitHub OAuth
- Microsoft OAuth
- Apple OAuth
- SAML

Each WorkOS connection costs us $125/mo, so this will be for Ent tiers only. Others can look into e.g. https://aglide.com.

Manual account setup right now, i.e. adding `sso_organization_id`, etc.:

```rb
account.update(
  sso_organization_domains: %w[example.com],
  sso_organization_id: 'org_123',
)
```

## Prereqs

- [ ] Write integration/unit tests for SSO.
- [ ] Write tests for session-based authentication.
- [ ] Third-party security review.

## Subreqs

- [ ] Update Keygen EE marketing pages to mention SSO.
- [ ] Add docs on new SSO-related error codes.
- [ ] Add docs on configuring SSO/SAML.
- [ ] Cut a new self-hosted release?
- [ ] Send Stdout 9.

### DIFF
--- a/app/controllers/auth/sso_controller.rb
+++ b/app/controllers/auth/sso_controller.rb
@@ -8,63 +8,27 @@ module Auth
 
     skip_verify_authorized
 
-    typed_query {
-      param :error_description, type: :string, optional: true, allow_blank: true
-      param :error, type: :string, optional: true, allow_blank: true
-      param :state, type: :string, optional: true, allow_blank: true
-      param :code, type: :string
-    }
     def callback
-      if (error_code = sso_query[:error]).present?
-        error_message = sso_query[:error_description]
+      Keygen::SSO.raise_on_request_error!(request) # handle errors
 
-        raise Keygen::Error::InvalidSingleSignOnError.new(
-          code: error_code.then { "SSO_#{_1.upcase}" unless _1.nil? },
-          detail: error_message,
-        )
-      end
+      code, state = request.query_parameters.values_at(:code, :state)
+      profile     = Keygen::SSO.redeem_code(code:)
+      account     = Keygen::SSO.lookup_account(
+        profile:,
+      )
 
-      code, state = sso_query.values_at(:code, :state)
-      profile     = WorkOS::SSO.profile_and_token(client_id: WORKOS_CLIENT_ID, code:)
-                               .profile
+      # assert the authn state matches what we expect e.g. the original email
+      # equals actual authenticated email
+      Keygen::SSO.raise_on_state_error!(state,
+        profile:,
+        account:,
+      )
 
-      account = Account.where.not(sso_organization_id: nil) # sanity-check
-                       .find_by(
-                         sso_organization_id: profile.organization_id,
-                       )
-
-      raise Keygen::Error::InvalidSingleSignOnError.new('account is invalid') if
-        account.nil?
-
-      # WorkOS recommends JIT-user-provisioning: https://workos.com/docs/sso/jit-provisioning
-      #
-      # 1. First, we attempt to find the user by their profile ID.
-      # 2. Next, we attempt to find the user by their email.
-      # 3. Otherwise, create a new user.
-      #
-      # Afterwards, we keep the user's profile up-to-date.
-      user = account.users.then do |users|
-        users.find_by(sso_profile_id: profile.id) || users.find_or_initialize_by(email: profile.email) do |u|
-          u.sso_profile_id    = profile.id
-          u.sso_connection_id = profile.connection_id
-          u.sso_idp_id        = profile.idp_id
-          u.first_name        = profile.first_name
-          u.last_name         = profile.last_name
-          u.email             = profile.email
-
-          # TODO(ezekg) eventually implement workos groups?
-          u.grant_role! :admin
-        end
-      end
-
+      user = Keygen::SSO.lookup_or_provision_user(profile:, account:)
       user.update!(
         # Generate a nonce to assert that only 1 SSO-based session can be
         # active for a user at any given time.
         session_nonce: SecureRandom.random_number(2**32),
-        # Keep the user's profile up-to-date with the IdP.
-        first_name: profile.first_name,
-        last_name: profile.last_name,
-        email: profile.email,
       )
 
       # We use encrypted session cookies for SSO authentication because we
@@ -74,13 +38,7 @@ module Auth
       session[:account_id] = account.id
       session[:user_id]    = user.id
 
-      redirect_to portal_url(account), status: :temporary_redirect,
-                                       allow_other_host: true
-    rescue WorkOS::APIError => e
-      raise Keygen::Error::InvalidSingleSignOnError.new(
-        code: e.code.then { "SSO_#{_1.upcase}" unless _1.nil? },
-        detail: e.message,
-      )
+      redirect_to portal_url(account), status: :see_other, allow_other_host: true
     end
   end
 end
--- a/app/controllers/auth/sso_controller.rb
+++ b/app/controllers/auth/sso_controller.rb
@@ -1,0 +1,49 @@
+# frozen_string_literal: true
+
+module Auth
+  KEYGEN_PORTAL_HOST = ENV.fetch('KEYGEN_PORTAL_HOST') { 'portal.keygen.sh' }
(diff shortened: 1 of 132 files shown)


app/controllers/concerns/authentication.rb
```diff
@@ -79,6 +117,23 @@ def authenticate_with_http_license(&auth_procedure)
     auth_procedure.call(license_key)
   end
 
+  def http_session_authenticator(session)
+    return nil if
+      current_account.nil? || session.nil?
+
+    user = current_account.users.for_environment(current_environment, strict: current_environment.nil?)
```
ezekg, line 124: Still need to set current account from session.

app/models/user.rb
```diff
@@ -470,6 +470,8 @@ def status
     end
   end
 
+  def single_sign_on_enabled? = account.sso?
```
ezekg, line 473: Disable updates to email when SSO is enabled.

ezekg: Should we also disable password reset fulfillment for accounts with SSO?

ezekg: Ref: https://workos.com/blog/the-developers-guide-to-sso

app/controllers/auth/sso_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+    def callback
+      code, state = sso_query.values_at(:code, :state)
+      profile     = WorkOS::SSO.profile_and_token(client_id: WORKOS_CLIENT_ID, code:)
+                               .profile
+
+      # TODO(ezekg) error handling e.g. code is invalid, error callback,
+      #             failure to find/create user, etc.
+
+      account = Account.find_by!(sso_organization_id: profile.organization_id)
+      user    = account.users.find_or_create_by!(email: profile.email) do |u|
```
ezekg, line 24: Should this do a find by `profile.connection_id` instead?

app/controllers/concerns/authentication.rb
```diff
@@ -97,42 +152,38 @@ def query_token_authenticator(query_token)
   end
 
   def http_password_authenticator(username = nil, password = nil)
+    if current_account.sso? && current_account.sso_organization_domains.any? { username.ends_with?(_1) } # for JIT-user-provisioning
+      redirect = Keygen::SSO.authorization_url(account: current_account, email: username)
```
ezekg, line 156: Need to pass in `provider:`.

app/controllers/concerns/authentication.rb
```diff
@@ -97,42 +152,38 @@ def query_token_authenticator(query_token)
...
-      raise Keygen::Error::UnauthorizedError.new(
-        detail: 'email and password must be valid',
-        code: 'CREDENTIALS_INVALID',
-        header: 'Authorization',
-      )
+      raise Keygen::Error::InvalidCredentialsError.new(header: 'Authorization')
+    end
+
+    if user.single_sign_on_enabled?
+      redirect = Keygen::SSO.authorization_url(account: current_account, email: user.email)
```
ezekg, line 169: Need to pass in `provider:`.

app/controllers/concerns/authentication.rb
```diff
@@ -97,42 +152,38 @@ def query_token_authenticator(query_token)
   end
 
   def http_password_authenticator(username = nil, password = nil)
+    if current_account.sso? && current_account.sso_organization_domains.any? { username.ends_with?(_1) } # for JIT-user-provisioning
```
ezekg, line 155: Is this the best way to check if the user's email is for one of the org's domains?

ezekg: For example, if the org domain was `actor.example`, `badactor.example` would also match.

config/application.rb
```diff
@@ -34,6 +34,17 @@ class Application < Rails::Application
     # Skip views, helpers and assets when generating a new resource.
     config.api_only = true
 
+    # Use cookies for sessions
+    config.session_store :cookie_store, key: '_keygen_session',
```
ezekg, line 38: Should we use a `Session` model instead? It'd give us auditable sessions and also make them revocable.

ezekg: We could start with a `UserSession` model.

ezekg: Ref https://github.com/rails/rails/pull/52328.

ezekg: Maybe we should do the session stuff in a separate PR?

app/controllers/auth/sso_controller.rb
```diff
@@ -0,0 +1,42 @@
+# frozen_string_literal: true
+
+module Auth
+  class SsoController < Api::V1::BaseController
+    include ActionController::Cookies
+
```
ezekg, line 6: This controller should assert the SSO entitlement, i.e. it should only be available in Keygen EE.

app/models/account.rb
```diff
@@ -289,6 +289,16 @@ def protected?
     protected
   end
 
+  def sso? = sso_organization_id?
```
ezekg, line 292: Should we also check `ent?`, or should we let this be open in case we sell an SSO add-on for Std tiers too?

lib/keygen/sso.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+module Keygen
+  module SSO
```
ezekg, line 4: Move to EE namespace.

app/controllers/concerns/authentication.rb
```diff
@@ -96,43 +157,42 @@ def query_token_authenticator(query_token)
...
-        raise Keygen::Error::UnauthorizedError.new(
-          detail: 'second factor must be valid',
-          code: 'OTP_INVALID',
-          pointer: '/meta/otp',
-        )
+        raise Keygen::Error::InvalidSecondFactorError.new(pointer: '/meta/otp')
       end
     end
 
     unless user.password? && user.authenticate(password)
```
ezekg, line 194: Should we return a `PASSWORD_REQUIRED` error code when SSO is not required, and the user has a password set but no password was provided? This could let us do the typical email first, password second, which allows SSO to replace password for some users (and eventually e.g. magic links, etc.)

ezekg:
Should we add a `/v1/sessions` route that aliases `/v1/tokens`, or should SSO/magic-links move to a sessions route?

PR #867, opened 2024-07-24
Update importer to validate account

### LINKED ISSUE #866: Importer doesn't validate record account IDs
This is a social engineering risk, where we import an account but it contains records (e.g. admin records) for another account. Right now, we manually verify the contents of the import, but I'd prefer to have these assertions enforced via code as well.

We should assert the following during import:

1. the account being imported does not already exist.
2. all records belong to the account being imported.
3. raise on duplicate?

For 3, we can use `insert_all!` instead of `insert_all`. This would make re-imports fail, but that seems OK.

ezekg:
Closes #866.

### DIFF
--- a/db/migrate/20240724204821_add_unique_resource_index_to_roles.rb
+++ b/db/migrate/20240724204821_add_unique_resource_index_to_roles.rb
@@ -1,0 +1,8 @@
+class AddUniqueResourceIndexToRoles < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+  verbose!
+
+  def change
+    add_index :roles, %i[resource_id resource_type], unique: true, algorithm: :concurrently, if_not_exists: true
+  end
+end
--- a/lib/keygen/importer/v1/importer.rb
+++ b/lib/keygen/importer/v1/importer.rb
@@ -7,7 +7,8 @@ module Keygen
   module Importer
     module V1
       class Importer
-        def initialize(secret_key: nil)
+        def initialize(account_id:, secret_key: nil)
+          @account_id   = account_id
           @deserializer = Deserializer.new(secret_key:)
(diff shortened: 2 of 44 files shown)


db/migrate/20240724204821_add_unique_resource_index_to_roles.rb
```diff
@@ -0,0 +1,8 @@
+class AddUniqueResourceIndexToRoles < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+  verbose!
+
+  def change
+    add_index :roles, %i[resource_id resource_type], unique: true, algorithm: :concurrently
```
ezekg, line 6: This is another line of defense for attempts at e.g. updating an existing user's role via import.

lib/keygen/importer/v1/importer.rb
```diff
@@ -21,18 +22,36 @@ def import(from:)
...
+          rescue ActiveRecord::RecordNotUnique
+            raise DuplicateRecordError
+          end
+        end
+
+        def validate_records!(records)
+          case records
+          in [Account => account] if account.id == account_id
+            # ok -- account is expected account
+          in [*] if records.all? { !_1.respond_to?(:account_id) || _1.account_id == account_id }
```
ezekg, line 50: We should adjust associations so that all exportable records have an account association, even if through associations.

spec/lib/keygen/importer_spec.rb
```diff
@@ -54,9 +54,9 @@
     let(:secret_key) { SecureRandom.hex }
 
     before do
-      create_list(:user, rand(1..100), account:)
```
ezekg, line 51: Removed randomness to speed up tests.

PR #886, opened 2024-09-12
Add docker compose

### LINKED ISSUE #727: Add Docker Compose support
I'm considering adding Keygen to our small business. I've just read the self-host documentation and, although the solution is based on Docker, there is no mention of how to make an optimal setup with Docker Compose. I.e., you have to have Postgres, Redis and stuff outside Docker or separately add those containers on your own.  
Would be nice to have a Compose example if you at Keygen use it yourselves.

neo773:
This PR adds one click `docker-compose` setup 

https://github.com/user-attachments/assets/4328cced-1f95-4175-bb44-2f4dafd4c34c

/claim #727
/closes #727

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -1,0 +1,294 @@
+# ------------------------------------------------------------------------------
+# Keygen is a twelve-factor app, and as such is configured through environment
+# variables. Uncommented environment variables are mandatory. Failure to
+# provide all mandatory environment variables will result in pain.
+#
+# For more information, visit: https://keygen.sh/docs/self-hosting
+# ------------------------------------------------------------------------------
+
+# The email used by Caddy for generating and managing Let's Encrypt TLS certificates.
+#
+# Leave blank in non-production environments and Caddy will generate a self-signed certificate, which can be trusted like so:
+#
+#   docker cp caddy:/data/caddy/pki/authorities/local/root.crt ./caddy-root.crt
+#   sudo cp ./caddy-root.crt /usr/local/share/ca-certificates/caddy-root.crt
+#   sudo update-ca-certificates
+#
+# In production environments, this is required to use Let's Encrypt.
+# CADDY_ACME_EMAIL=
+
+# The username for the PostgreSQL database.
+POSTGRES_USER=
+
+# The password for the PostgreSQL database.
+POSTGRES_PASSWORD=
+
+# The name of the PostgreSQL database.
+POSTGRES_DB=
+
+# The Redis database URL, e.g. redis://localhost:6379.
+REDIS_URL=redis://redis:6379
+
+# An internal secret key used by Ruby on Rails.
+# Use `rails secret` or `openssl rand -hex 64` to generate one. We recommend a random 64-byte secret.
+SECRET_KEY_BASE=
+
+# The deterministic key used for at-work encryption.
+# Use `rails db:encryption:init` or `openssl rand -base64 32` to generate one. We recommend a random 32-byte secret. No less than 12-bytes.
+ENCRYPTION_DETERMINISTIC_KEY=
+
+# The primary key used for at-work encryption.
+# Use `rails db:encryption:init` or `openssl rand -base64 32` to generate one. We recommend a random 32-byte secret. No less than 12-bytes.
+ENCRYPTION_PRIMARY_KEY=
+
+# The key derivation salt used for at-work encryption.
+# Use `rails db:encryption:init` or `openssl rand -base64 32` to generate one. We recommend a random 32-byte secret. No less than 20-bytes.
+ENCRYPTION_KEY_DERIVATION_SALT=
+
+# The edition of Keygen to run: either CE for the Community Edition, or EE for the Enterprise Edition.
+KEYGEN_EDITION=CE
+
+# The mode of Keygen: either singleplayer for single-tenant, or multiplayer for multi-tenant.
+# Multiplayer mode requires a Keygen EE license with the multiplayer entitlement.
+KEYGEN_MODE=singleplayer
+
+# The account ID used for setting the tenant of Keygen in singleplayer mode.
+# Use `uuidgen` or `cat /proc/sys/kernel/random/uuid` to set a pre-determined ID.
+# Required in singleplayer mode.
+KEYGEN_ACCOUNT_ID=
+
+# The admin email used during setup. Must be a valid email address for your organization.
+KEYGEN_ADMIN_EMAIL=
+
+# The admin password used during setup. Must be at least 6 characters.
+# Use `openssl rand -hex 16` to generate one. We recommend a random 16-byte secret.
+KEYGEN_ADMIN_PASSWORD=
+
+# The primary host that Keygen's API will be accessible at, e.g. licensing.example.com.
+KEYGEN_HOST=
+
+# Additional hosts that Keygen will be accessible at.
+# Must be a comma-delimited list of hostnames, e.g. host1.example.com,host2.example.com.
+# KEYGEN_HOSTS=
+
+# The path to a Keygen EE license file, e.g. /etc/keygen/ee.lic.
+# Either KEYGEN_LICENSE_FILE_PATH or KEYGEN_LICENSE_FILE are required in Keygen EE.
+# KEYGEN_LICENSE_FILE_PATH=
+
+# A base64 encoded Keygen EE license file. This is useful when a file system is not available for KEYGEN_LICENSE_FILE_PATH, e.g. with Heroku.
+# Either KEYGEN_LICENSE_FILE_PATH or KEYGEN_LICENSE_FILE are required in Keygen EE.
+# KEYGEN_LICENSE_FILE=
+
+# The license key used to decrypt the Keygen EE license file.
+# Required in Keygen EE.
+# KEYGEN_LICENSE_KEY=
+
+# The primary domain Keygen's API will be accessible at. By default, this is parsed from KEYGEN_HOST, but that may result in inaccuracies under certain TLDs, e.g. co.uk.
+# KEYGEN_DOMAIN=
+
+# The primary subdomain Keygen's API will be accessible at. By default, parsed from KEYGEN_HOST, but that may result in inaccuracies under multiple subdomains.
+# KEYGEN_SUBDOMAIN=
+
+# The number of days that high-volume event logs are kept before being pruned.
+# Set to -1 to disable pruning.
+# KEYGEN_PRUNE_EVENT_BACKLOG_DAYS=90
+
+# The number of days that request logs are kept before being pruned.
+# Set to -1 to disable pruning.
+# KEYGEN_PRUNE_REQUEST_BACKLOG_DAYS=30
+
+# The number of days that webhook events are kept before being pruned.
+# Set to -1 to disable pruning.
+# KEYGEN_PRUNE_WEBHOOK_BACKLOG_DAYS=30
+
+# The number of rows pruned per-batch during pruning.
+# KEYGEN_PRUNE_BATCH_SIZE=10000
+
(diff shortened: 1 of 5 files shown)


scripts/docker-compose.sh
```diff
@@ -0,0 +1,69 @@
...
+echo "Waiting for PostgreSQL to be ready..."
+until docker-compose exec -T postgres pg_isready; do
+  echo "PostgreSQL is unavailable - sleeping"
+  sleep 1
+done
+echo "PostgreSQL is up and running"
+
+# Run setup service
+echo "Running setup service..."
+docker-compose run --rm setup
```
ezekg, line 60: Is there a way to make it so that the setup doesn't run every time the script is run? The `setup` script runs `rails db:schema:load`, which wipes the database, i.e. a dangerous action for a "turn-key" setup like Compose.

The `setup` command is only meant to be run once during initial install, to create the account and admin user. This is why in https://github.com/keygen-sh/keygen-api/issues/727#issuecomment-1590322284, I suggested maybe we skip the `setup` command entirely for Compose, and use seeds instead,  which are safe to run multiple times (and recommended to run on each deploy to apply e.g. new permissions and event types).

Hope that makes sense? wdyt?

ezekg (changes requested):
Had a suggestion on overall direction. Also, can you add better instructions to the PR description? I assume this is meant to be run via: `./scripts/docker-compose.sh` and not `docker-compose -f docker-compose.yml up -d`? Is this typical? I'd like to follow what other projects do for consistency. I don't use Compose enough to have an opinion, but this seems atypical to what I've seen with other Compose implementations for self-hosting.

ezekg:
I would prefer to do away with the `docker-compose.sh` script, and do something more along the lines of what Chatwoot is doing: https://www.chatwoot.com/docs/self-hosted/deployment/docker/, e.g. as part of initial setup, have them run:

```sh
docker compose run -i --rm web bundle exec rails keygen:setup
```

I also want to do away with all hard-coded passwords in `.env.sample`. This is just asking for somebody to setup a production database using defaults. I'm fine with requiring somebody manually setup their `.env` file i.e. not doing it for them.

And if we're going to go the route of a `.env.sample`, we need to actually cover [all required environment vars](https://keygen.sh/docs/self-hosting/#configuration). Let's mirror what [Chatwoot does here](https://github.com/chatwoot/chatwoot/blob/develop/.env.example)? Seems pretty straight forward and reduces "magic" unlike the current shell script.

We should also not have a default for `KEYGEN_HOST`, since that must be set to the actual domain name being used — a default is just going to cause more trouble than it helps.

The [message after `setup`](https://github.com/keygen-sh/keygen-api/blob/master/lib/tasks/keygen/setup.rake#L107-L110) can be adjusted to mention adding vars to the current shell environment, or to `.env`.

Then we can also instruct migrations and seeds to be run before each deployment, similarly to the `setup`:

```sh
docker compose run --rm web bundle exec rails db:migrate db:seed
```

Overall, I want to avoid footguns here — people using Compose want turn-key.

wdyt?

neo773:
Sounds good, I initially went with 1 click setup as I wanted it to be as frictionless as possible.

Also good idea to cover all ENV vars will do that.

.env.sample
```diff
@@ -1,8 +1,280 @@
-POSTGRES_USER=postgres
-POSTGRES_PASSWORD=postgres
-POSTGRES_DB=keygen
-KEYGEN_HOST=api.keygen.localhost
-KEYGEN_EDITION=CE
-KEYGEN_MODE=singleplayer
-KEYGEN_ADMIN_EMAIL=test@test.com
-KEYGEN_ADMIN_PASSWORD=123456
+# ------------------------------------------------------------------------------
+# Environment Variables
```
ezekg, line 2: Can you add a link to the self-hosting docs here in case there are additional questions/concerns?

.env.sample
```diff
@@ -0,0 +1,280 @@
...
+# The primary key used for at-work encryption.
+# Use `rails db:encryption:init` or `openssl rand -base64 32` to generate one. We recommend a random 32-byte secret. No less than 12-bytes.
+ENCRYPTION_PRIMARY_KEY=
+
+# The key derivation salt used for at-work encryption.
+# Use `rails db:encryption:init` or `openssl rand -base64 32` to generate one. We recommend a random 32-byte secret. No less than 20-bytes.
+ENCRYPTION_KEY_DERIVATION_SALT=
+
+# The edition of Keygen to run: either CE for the Community Edition, or EE for the Enterprise Edition.
+KEYGEN_EDITION=
```
ezekg, line 33: Fine to default to `CE` here.

.env.sample
```diff
@@ -0,0 +1,280 @@
...
+
+# The mode of Keygen: either singleplayer for single-tenant, or multiplayer for multi-tenant.
+# Multiplayer mode requires a Keygen EE license with the multiplayer entitlement.
+KEYGEN_MODE=
```
ezekg, line 37: Fine to default to `singleplayer` here.

.env.sample
```diff
@@ -0,0 +1,280 @@
...
+
+# The account ID used for setting the tenant of Keygen in singleplayer mode.
+# Use `bundle exec rails keygen:setup` to generate one, or use `uuidgen` or `cat /proc/sys/kernel/random/uuid` to set a pre-determined ID.
```
ezekg, line 40: Can we remove the bit about `bundle exec rails keygen:setup`?

.env.sample
```diff
@@ -0,0 +1,280 @@
...
+# The number of database connections Keygen will use per-process.
+# By default, this will equal RAILS_MAX_THREADS or SIDEKIQ_CONCURRENCY, determined by web vs worker process, respectively (i.e. each thread will have its own database connection).
+# Because of this default, most applications do not need to set this. But it's there if you need to override the default behavior for some reason.
+# DB_POOL=
+
+# The database query timeout, in ms.
+# DB_TIMEOUT=5000
+
+# The Redis database URL, e.g. redis://localhost:6379.
+# REDIS_URL=
```
ezekg, line 112: Redis is required.

neo773:
To set it up you'd run it like this.

```
docker-compose --profile setup run --rm setup

docker compose up
```

neo773:
Pushed and squashed

ezekg:
Could we update [this message](https://github.com/keygen-sh/keygen-api/blob/master/lib/tasks/keygen/setup.rake#L107-L110) to something along the lines of this?

```
To complete setup, add the following env vars in your shell or to .env:

  SECRET_KEY_BASE=xxx
  ENCRYPTION_DETERMINISTIC_KEY=xxx
  ENCRYPTION_PRIMARY_KEY=xxx
  ENCRYPTION_KEY_DERIVATION_SALT=xxx
  KEYGEN_EDITION=CE
  KEYGEN_MODE=singleplayer
  KEYGEN_HOST=licensing.example.com
  KEYGEN_ACCOUNT_ID=xxx

In addition, you may want to keep track of the following information:

  Account ID: xxx
  Account slug: me
  Admin email: me@example.com

Then start the web and worker processes.

*keygen music intensifies*
```

neo773:
@ezekg 
Shouldn't we display it if user hasn't set them?

ezekg:
> @ezekg Shouldn't we display it if user hasn't set them?

It's fine to display all vars. If the user has already set them, the values won't be overridden so it will just print current vals.

ezekg:
Also —one final ask: is there a way to add Caddy to this to automatically handle TLS and reverse-proxy side of things? 🙂

Ideally, it'd serve all domains in `KEYGEN_HOST` and `KEYGEN_HOSTS`.

neo773:
> Also —one final ask: is there a way to add Caddy into this to automatically handle TLS and reverse-proxy side of things? 🙂
> 
> Ideally, it'd serve all domains in `KEYGEN_HOST` and `KEYGEN_HOSTS`.

I'd figured since we're keeping it less opinionated it might be better to leave that off to user?
They could use reverse proxy of their choice which could Nginx or Caddy etc.

Let me know what you think.

ezekg:
> I'd figured since we're keeping it less opinionated it might be better off to leave that to user? They could use reverse proxy of their choice which could Nginx or Caddy etc.
> 
> Let me know what you think.

I think it's fine to have Caddy be the default here. People can always modify the compose file to use Nginx or Apache. But Caddy gives the 99% an easy way to get a reverse-proxy and TLS set up with minimal effort. This is a big pain point for people not super knowledgeable in networking when it comes to self-hosting Keygen CE.

I'm all for doing things the Rails way — sensible defaults, config when needed.

> > To complete setup, add the following env vars in your shell or to .env:
> 
> Looks like current file is already printing all the variables however we could change this part
> 
> ```
> Then run the following to start the server:
> 
> rails s
> ```
> 
> to if we're inside docker container set by `IS_DOCKER_SETUP` env var in the config
> 
> ```
> Then run the following to start the server:
> 
> docker-compose up
> ```

I want to change it to a more general: "Then start the web and worker processes." as I mentioned above.

docker-compose.yaml
```diff
@@ -0,0 +1,106 @@
...
+      - "80:80"
+      - "443:443"
+    volumes:
+      - caddy_data:/data
+      - caddy_config:/config
+    env_file:
+      - .env
+    command: >
+      sh -c '
+        echo "${KEYGEN_HOST} ${KEYGEN_HOSTS} {
```
ezekg, line 85: Will this work even though `KEYGEN_HOSTS` is a comma-delimited list?

ezekg: Fails for me:

```
caddy-1     | {"level":"info","ts":1726262644.073456,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}                                                                                                                                                                                                         
caddy-1     | Error: adapting config using caddyfile: Site addresses cannot contain a comma ',': 'auth.keygen.localhost,get.keygen.localhost' - put a space after the comma to separate site addresses, at /etc/caddy/Caddyfile:1
caddy-1     | {"level":"info","ts":1726262647.809099,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}                                                                                                                                                                                                         
caddy-1     | Error: adapting config using caddyfile: Site addresses cannot contain a comma ',': 'auth.keygen.localhost,get.keygen.localhost' - put a space after the comma to separate site addresses, at /etc/caddy/Caddyfile:1
caddy-1     | {"level":"info","ts":1726262654.78514,"msg":"using config from file","file":"/etc/caddy/Caddyfile"}                                                                                                                                                                                                          
caddy-1     | Error: adapting config using caddyfile: Site addresses cannot contain a comma ',': 'auth.keygen.localhost,get.keygen.localhost' - put a space after the comma to separate site addresses, at /etc/caddy/Caddyfile:1
```

ezekg: You can use variable expansion, `${KEYGEN_HOSTS/,/ }`.

ezekg: Just thinking out loud here — would it be cleaner to just add an actual `Caddyfile` to the repo? That way people can easily customize it if needed. I think Caddy supports env vars in the Caddyfile.

neo773: KEYGEN_HOSTS is a blocker tho, it throws the same error (put a space after the comma to separate site addresses).

I got with working with shell script

neo773:
```    command: >
      sh -c '
        echo "${KEYGEN_HOST}$${KEYGEN_HOSTS:+,} $${KEYGEN_HOSTS//,/, } {
          tls internal
          reverse_proxy web:3000
        }" > /etc/caddy/Caddyfile &&
        caddy run --config /etc/caddy/Caddyfile
      '```

ezekg:
@neo773 I updated the Caddy setup. I wanted to make it more production-capable out-of-the-box.

wdyt?

ezekg (approved):
/approve

PR #894, opened 2024-10-07
Add support for raw engine

### LINKED ISSUE #892: Add raw distribution engine
Right now, the current `get.keygen.sh` subdomain doesn't really support multiple products per-account, or at least makes it hard to do so. Let's take us as an example — we have the Keygen CLI, and now we have Keygen Relay.

Versions for the CLI and Relay now conflict for a request like `get.keygen.sh/keygen/1.0.0/cli_linux_386`, since there are now multiple `1.0.0` releases — a CLI release, and a Relay release, and this URL will automatically scope to the newer release. So even when the artifact exists, it may respond with a 404 if it's scoped to the wrong `1.0.0` release i.e. Relay vs CLI.

A way forward here would be to create a "raw" engine that includes account, product, package, and release information in the URL. This way download URLs can be super specific i.r.t. scoping.

Thinking this URL pattern, making the `@<package>` scope optional ala npm:

```
GET https://raw.pkg.keygen.sh/<account>/<product>[/@<package>]/<release>/<artifact>
```

See #679 for implementation of the PyPI engine, and #753 for the Tauri engine; the Tauri one will likely be the most helpful for reference, since the PyPI included a lot of noisy ground work. But both can be good reference material.

Depends on #893.

ezekg:
Closes #892.


app/models/release_artifact.rb
```diff
@@ -173,7 +173,16 @@ class ReleaseArtifact < ApplicationRecord
...
+    case product
+    in UUID_RE => product_id
+      joins(:product).where(product: { id: product_id })
+    in Product => product
+      joins(:product).where(product:)
+    in String => code
+      joins(:product).where(product: { code: })
+    else
+      none
+    end
```
ezekg, line 185: Todo: other `for_product` scopes should accept a `code` as well.

app/controllers/api/v1/products_controller.rb
```diff
@@ -136,9 +138,11 @@ def destroy
     def set_product
       scoped_products = authorized_scope(current_account.products)
 
-      @product = scoped_products.find(params[:id])
-
-      Current.resource = product
+      Current.resource = @product = FindByAliasService.call(
```
ezekg, line 141: I'd like to eventually move this into a `find_by_alias` concern, where a model can define aliases:

```rb
class Product
  include Aliasable
  
  has_aliases :code
end
```

PR #895, opened 2024-10-08
feat: support custom AWS S3 Endpoint

aminya:
This PR allows setting the AWS S3 endpoint URL. This allows for using S3-compatible services like Tigris or self-hosting the storage.

https://www.tigrisdata.com/docs/sdks/s3/aws-ruby-sdk/


config/initializers/s3.rb
```diff
@@ -6,9 +6,11 @@
 AWS_SECRET_ACCESS_KEY = ENV['AWS_SECRET_ACCESS_KEY']
 AWS_BUCKET            = ENV['AWS_BUCKET']
 AWS_REGION            = ENV['AWS_REGION']
+AWS_ENDPOINT_URL_S3   = ENV['AWS_ENDPOINT_URL_S3']
 
 AWS_CLIENT_OPTIONS = {
   access_key_id: AWS_ACCESS_KEY_ID,
   secret_access_key: AWS_SECRET_ACCESS_KEY,
   region: AWS_REGION,
+  endpoint: AWS_ENDPOINT_URL_S3
```
ezekg, line 15: Does this break the default case where `endpoint: nil`? You may need to only set this key the value is non-nil.

ezekg:
Can you add some background here? What's the use case? Why is this valuable to you?

config/initializers/s3.rb
aminya: I didn't test the default configuration, but with specifying it, it works with no issues

aminya:
> Can you add some background here? What's the use case? Why is this valuable to you?

This allows using S3-compatible services like Tigris or self-host the storage.

https://www.tigrisdata.com/docs/sdks/s3/aws-ruby-sdk/

config/initializers/s3.rb
aminya: Would this be enough to support the default config?
```suggestion
  endpoint: AWS_ENDPOINT_URL_S3 or nil
```

ezekg: No, because that does nothing since `AWS_ENDPOINT_URL_S3` is already nil if it's not specified.

Right now, I get the following error when `endpoint: nil` when initializing the client:

```
error: missing required option `:endpoint' (ArgumentError)
```

Meaning, this will break the majority of Keygen installs, since this would be a new option.

This new feature needs to be implemented in a backwards compatible way. The above error indicates that `endpoint:` should only be set when present, since it's not allowed to be `nil`. Easiest way to do that would be to `compact` the `AWS_CLIENT_OPTIONS` hash before freezing it, which would remove any `nil` values.

aminya: I have addressed this in [`c5a45a3` (#895)](https://github.com/keygen-sh/keygen-api/pull/895/commits/c5a45a3a41ba2a2461755df85e6e7e1714c434e5)

config/initializers/s3.rb
```diff
@@ -6,9 +6,11 @@
...
 AWS_REGION            = ENV['AWS_REGION']
+AWS_ENDPOINT_URL_S3   = ENV['AWS_ENDPOINT_URL_S3']
 
 AWS_CLIENT_OPTIONS = {
   access_key_id: AWS_ACCESS_KEY_ID,
   secret_access_key: AWS_SECRET_ACCESS_KEY,
   region: AWS_REGION,
-}.freeze
+  endpoint: AWS_ENDPOINT_URL_S3
+}.compact!.freeze
```
ezekg, line 16:
```suggestion
}.compact.freeze
```

ezekg: The bang variant is not really needed here.

config/initializers/s3.rb
```diff
@@ -6,9 +6,11 @@
 AWS_SECRET_ACCESS_KEY = ENV['AWS_SECRET_ACCESS_KEY']
 AWS_BUCKET            = ENV['AWS_BUCKET']
 AWS_REGION            = ENV['AWS_REGION']
+AWS_ENDPOINT_URL_S3   = ENV['AWS_ENDPOINT_URL_S3']
```
ezekg, line 9: Is this a standardized name? I'd rather use `AWS_ENDPOINT_URL` if not, since none of the other keys have a `_S3` suffix.

hrueger:
Hi all,
we're also very interested in this. We've tried this PR, however, we have the problem that it automatically appends the bucket name as a subdomain layer:
`storage.ourserver.com` where we host a minio becomes `keygen-bucket.storage.ourserver.com`... Any idea on why that happens?

config/initializers/s3.rb
aminya: Yes, that's standard. Removing S3 causes conflicts with other endpoints AWS provides

https://fly.io/docs/tigris/#create-and-manage-a-tigris-storage-bucket

aminya:
> Hi all, we're also very interested in this. We've tried this PR, however, we have the problem that it automatically appends the bucket name as a subdomain layer: `storage.ourserver.com` where we host a minio becomes `keygen-bucket.storage.ourserver.com`... For Minio it would have to be storage.ourserver.com/keygen-bucket` though. Any idea on why that happens?

It seems Minio isn't following the standard convention. It works with Tigris

lukas-runge:
> Hi all, we're also very interested in this. We've tried this PR, however, we have the problem that it automatically appends the bucket name as a subdomain layer: `storage.ourserver.com` where we host a minio becomes `keygen-bucket.storage.ourserver.com`... For Minio it would have to be storage.ourserver.com/keygen-bucket` though. Any idea on why that happens?

Using the flag `force_path_style: true` in the client config fixes this. @aminya maybe add the option to set this flag via env variable. 🙌

config/initializers/s3.rb
aminya, line 9:
```suggestion
AWS_ENDPOINT_URL_S3   = ENV['AWS_ENDPOINT_URL_S3']
AWS_FORCE_PATH_STYLE_S3 = ENV['AWS_FORCE_PATH_STYLE_S3']
```

config/initializers/s3.rb
```diff
@@ -6,9 +6,11 @@
...
 
 AWS_CLIENT_OPTIONS = {
   access_key_id: AWS_ACCESS_KEY_ID,
   secret_access_key: AWS_SECRET_ACCESS_KEY,
   region: AWS_REGION,
-}.freeze
+  endpoint: AWS_ENDPOINT_URL_S3
```
aminya, line 15:
```suggestion
  endpoint: AWS_ENDPOINT_URL_S3,
  force_path_style: AWS_FORCE_PATH_STYLE_S3,
```

aminya:
@lukas-runge Great, does this cover your usecase?
[`f9d387c` (#895)](https://github.com/keygen-sh/keygen-api/pull/895/commits/f9d387cc775e73030b862abed34bd45260e4670a)

lukas-runge:
> @lukas-runge Great, does this cover your usecase? [`f9d387c` (#895)](https://github.com/keygen-sh/keygen-api/pull/895/commits/f9d387cc775e73030b862abed34bd45260e4670a)

Yes - just tested it. 👍

Ty @aminya! 🙏

ezekg (changes requested):
First off, I appreciate the PR, and the discussion from everybody here.

As-is this PR works, but the implementation is not ideal, because it doesn't take into account existing users wanting to migrate from one storage provider to another, e.g. from S3 to Tigris. What I mean is that if an existing Keygen CE user with artifacts already stored in S3 wanted to move from S3 to Tigris, they couldn't do so safely without replicating their entire S3 bucket into Tigris, which can be very costly for large objects. Without the replication, existing artifacts would no longer be accessible, since they would exist in S3 but not Tigris, and the S3 backend now points to Tigris and not S3 (which is confusing unto itself when there is an R2 backend that is also S3-compatible).

In the case of Keygen Cloud, when we moved from S3 to R2, we did so seamlessly, without replicating TBs of data from S3 to R2. We were able to do this because existing artifacts remained in S3, while new artifacts were stored in R2. And this is still the case to this day. I want the same story for Keygen CE/EE users moving from S3 or R2 to Tigris or Minio.

So with that said —

Rather than introduce customization/overloads for the S3 storage backend, e.g. `endpoint` and `force_path_style` (a config that the user has to understand), I'd rather we introduce dedicated Tigris and Minio backends. The Minio client can set `force_path_style: true`, so that it works out of the box without the user needing to know about that config.

Both new backends could be configured via `TIGRIS_` and `MINIO_` env vars, just like `AWS_` and `CF_` are used to configure S3 and R2, respectively. The code is already designed in a way [to be extensible](https://github.com/keygen-sh/keygen-api/blob/fb819b5b627f984387a236e3da79eb1597af3ead/app/models/release_artifact.rb#L466-L482), to be able to add `TIGRIS` and `MINIO`, and even more storage providers in the future. Each backend would have its own specific environment variables for configuring the client so that it "Just Works."

In addition to the new Tigris and Minio backends, we should add a `KEYGEN_STORAGE_BACKEND` environment variable to set [the default storage backend](https://github.com/keygen-sh/keygen-api/blob/fb819b5b627f984387a236e3da79eb1597af3ead/app/models/account.rb#L82) for new accounts, since there would be more than just the current S3 and R2 backends.

Migrating an existing account would be as simple as running:

```rb
account = Account.sole

account.update!(backend: 'TIGRIS')
```

Existing artifacts e.g. in R2 would still be accessible, since artifacts have their own `backend`, and new artifacts would go to Tigris, no replication needed.

Would you be open to adjusting the PR?

aminya:
The idea of having separate classes for each provider sounds good at the first glance, however it doesn't match the original promise of these services. As you see, Tigris simply recommends providing the endpoint, and it uses the same AWS environment variables. When you set up a Fly.io Tigris machine, the exact environment variables I used in this code are set automatically by Fly. So if we change these names, it's gonna make the deployment harder not easier.

https://www.tigrisdata.com/docs/sdks/s3/aws-ruby-sdk/

Another thing is that we can't simply find all the S3 providers in the world and provide wrappers for them. For example, it's my first time hearing about Minio.

ezekg:
I understand that's how S3-compatible services work, but what I want to avoid here is env vars like `AWS_FORCE_PATH_STYLE_S3` and others like it. It's easier to add support for backends like Minio (which I'm also unfamiliar with) that have the `force_path_style` baked-in so that the end-user doesn't have to think about it. Like I said, a proper backend is best, because although something may be 'S3-compatible', that doesn't mean it's compatible with the current client options we have for the S3 backend.

I'm fine with introducing a single `AWS_ENDPOINT` env var (you can try `AWS_ENDPOINT` first and then fallback to `AWS_ENDPOINT_URL_S3`), for general compatibility with S3-compatible object stores in the simplest case, but I don't want to introduce anything else configurable via the environment. And like I said, I'd love to see a PR adding proper a Tigris backend that users can migration to and from, in a backwards-compatible way, just like the current S3 and R2 backends.

What I'm not saying is to add support for every S3-compatible object store off-the-bat, otherwise we would've done that already. I'm fine with only supporting the backends that people actually want and use. We can add support as-needed — like with you adding support for Tigris, and perhaps @lukas-runge or somebody else adding support for Minio.

For the simplest case, I'm fine with something like this:

```rb
AWS_ACCESS_KEY_ID     = ENV['AWS_ACCESS_KEY_ID']
AWS_SECRET_ACCESS_KEY = ENV['AWS_SECRET_ACCESS_KEY']
AWS_BUCKET            = ENV['AWS_BUCKET']
AWS_REGION            = ENV['AWS_REGION']
AWS_ENDPOINT          = ENV['AWS_ENDPOINT'] || ENV['AWS_ENDPOINT_URL_S3']

AWS_CLIENT_OPTIONS = {
  access_key_id: AWS_ACCESS_KEY_ID,
  secret_access_key: AWS_SECRET_ACCESS_KEY,
  region: AWS_REGION,
  endpoint: AWS_ENDPOINT,
}.compact.freeze
```

We (not necessarily you) can do proper support for Tigris and Minio in separate PRs.

ezekg:
Quick follow up. I changed my mind.

After looking at what [other](https://github.com/gomods/athens/blob/531dcf6ff349778dd691e45534c25982d806bd2b/pkg/config/s3.go#L14) [open](https://github.com/transloadit/uppy/blob/8f97fd61fada0dee6fe32b07516568f8c3d39060/examples/aws-companion/server.cjs#L38) [source](https://github.com/erxes/erxes/blob/9f0af702dbbbf3f56f6fe4792ac7a53cd3b2ad7c/packages/workers/src/data/utils.ts#L88) [projects](https://github.com/taskcluster/taskcluster/blob/1f02e08d3ac9520fd636409dd44f97b9945bf882/docker/env/.queue#L32-L33) [do](https://github.com/ledermann/docker-rails/blob/develop/config/initializers/shrine.rb#L32-L33), I'm fine with introducing `AWS_FORCE_PATH_STYLE` as well:

```rb
AWS_ACCESS_KEY_ID     = ENV['AWS_ACCESS_KEY_ID']
AWS_SECRET_ACCESS_KEY = ENV['AWS_SECRET_ACCESS_KEY']
AWS_BUCKET            = ENV['AWS_BUCKET']
AWS_REGION            = ENV['AWS_REGION']
AWS_ENDPOINT          = ENV['AWS_ENDPOINT'] || ENV['AWS_ENDPOINT_URL_S3']
AWS_FORCE_PATH_STYLE  = ENV['AWS_FORCE_PATH_STYLE'] == 'true'

AWS_CLIENT_OPTIONS = {
  access_key_id: AWS_ACCESS_KEY_ID,
  secret_access_key: AWS_SECRET_ACCESS_KEY,
  region: AWS_REGION,
  endpoint: AWS_ENDPOINT,
  force_path_style: AWS_FORCE_PATH_STYLE,
}.compact_blank
 .freeze
```

For users that want proper support for a Tigris/Minio/etc. backend, that can be implemented in a separate PR.

aminya:
@ezekg Changed the variable names accordingly.

ezekg (approved):

ezekg:
CLA [has already been signed](https://cla-assistant.io/keygen-sh/keygen-api?pullRequest=895) per the GH workflow. Been a busy week with other releases (Relay, OCI engine, etc.). But will merge this soon and it'll go out in the next self-hosted release. Really appreciate the PR.

ezekg:
Merged. Will go out in the next self-hosted release. Since we just cut a CE/EE release last week, it may be awhile before we cut a new release, since our self-hosted release cadence is typically bi-annual. You may need to build the image from master if you need to use the new features earlier, but just understand the risks associated with running edge software.

Issue #491, opened 2021-11-04
Add a release engine for Rubygems


ezekg:
This is part of our overarching goal of supporting more package managers in Keygen: https://github.com/keygen-sh/keygen-api/issues/493.

See #679 for implementation of PyPI, and #753 for Tauri. Bulk of the PyPI logic is in [the controller](https://github.com/keygen-sh/keygen-api/blob/master/app/controllers/api/v1/release_engines/pypi/simple_controller.rb), with integration tests [here](https://github.com/keygen-sh/keygen-api/blob/master/features/api/v1/engines/pypi/simple/index.feature) and [here](https://github.com/keygen-sh/keygen-api/blob/master/features/api/v1/engines/pypi/simple/show.feature). The Tauri implementation can be found in the aforementioned PR, since it's much simpler than the PyPI PR, as the PR for PyPI contained a lot of ground work for future packaging systems and may not be particularly helpful.

Implementation-wise here, we need to implement the expected [API for Rubygems](https://github.com/KN4CK3R/gitea/blob/4b1f387cb90e15d0f1659b96df6a283f98c65438/routers/api/packages/api.go#L368-L378), in a way that's compatible with Keygen's existing distribution schema. May be able to use [the Rubygems lib](https://github.com/rubygems/rubygems) for heavy lifting here, but I'm not sure. Like PyPI and Tauri, the implementation should only be concerned with downloading gems, not uploading. Drafting releases and uploading artifacts to Keygen's distribution API can be done via [Keygen's CLI](https://keygen.sh/docs/cli/), or via the Dashboard.

For an overview on the API endpoints expected by Rubygems, check out this blog post: https://keygen.sh/blog/how-to-license-and-distribute-a-private-rubygem/. We can use that to model the Rubygems engine here. This engine is meant to replace the workflow in that blog post — it should be a turn-key way to distribute gems.

In the end, I want a customer to be able to upload a `.gem` file (via `gem build`) to the distribution API we already have, and allow their end-users to `gem install` that gem like they would from https://rubygems.org.

Implementation should include [integration tests](https://github.com/keygen-sh/keygen-api/blob/master/features/api/v1/engines/tauri/v2/upgrade.feature). lmk of any questions.

mobley-trent:
Currently configuring my environment... Will it be easier if docker-compose support is integreted first @ezekg ?

ezekg:
@mobley-trent We're working on that in #727 but no ETA yet.

[...]

Issue #489, opened 2021-11-04
Add a release engine for npm


ezekg:
This is part of our overarching goal of supporting more package managers in Keygen: https://github.com/keygen-sh/keygen-api/issues/493.

Implement similar to #478. I think I’d prefer a `node` or `npm` subdomain. See #679 for implementation of PyPI, and #753 for 
Tauri. The Tauri one will likely be the most helpful for reference, since the PyPI included a lot of noisy ground work.

Helpful resources:

- https://keygen.sh/blog/how-to-license-and-distribute-commercial-node-modules/
- https://github.com/keygen-sh/example-private-npm-package

This engine is meant to replace the workflow in that blog post — it should be a turn-key way to distribute modules.

In the end, I want a customer to be able to upload a module (e.g. via `npm pack`) to the distribution API we already have, and allow their end-users to `npm install` that module like they would from https://npmjs.com.

Implementation should include [integration tests](https://github.com/keygen-sh/keygen-api/blob/master/features/api/v1/engines/tauri/v2/upgrade.feature). lmk of any questions.

ezekg:
What makes this hard is that npm's [package abbreviated manifest](https://github.com/npm/registry/blob/master/docs/responses/package-metadata.md#abbreviated-metadata-format) requires `dependency` data, and I don't want to add one-off functionality to the API to treat `package.json` artifacts as "special."

The simplest route here would be to require additional fields like `dependencies`, `devDependencies`, `peerDependencies`, `bin`, etc. be supplied in the `package.json` artifact's metadata. This approach aligns with what we do for PyPI artifact metadata as well, e.g. `data-requires-python`.

ezekg:
Helpful article on the API expected by npm and friends: https://keygen.sh/blog/how-to-license-and-distribute-commercial-node-modules/. We can use that to model the npm engine here.

rajeshj11:
@varshith257 I would like to collaborate with you. If you're interested, please let me know how I can assist.

ezekg:
Thought it may be helpful to see how others do it: https://github.com/KN4CK3R/gitea/blob/4b1f387cb90e15d0f1659b96df6a283f98c65438/routers/api/packages/api.go#L304-L348

ezekg:
[...]

PR #896, opened 2024-10-08
Add Rubygems release engine.

nashby:
@ezekg here's my a bit more than a POC of the Rubygems release engine. Would be great if you could take a look and see if that's something you like where it's going.

A couple of notes/questions:

- currently all the code is in controller. Even though it's a bit "fat" I think it still makes sense to keep it there without extracting it to some service. On the other hand it makes sense to extract it since it would be easier to write unit (acceptance?) tests for it but as far as I can see this project is mostly covered by system tests (cucumber features). Do you have a strong opinion about it?
- even though I've been testing it both with installing gem by `gem install` and by `bundle install` I had to comment out all authorization code to make it work since I was having issue with making it work with my data from seeds. Any tips on how can I ensure I have a list of releases that accessible by some user?
- how do you test artifacts in dev/test envs if this project uses S3 only without any fallback to filesystem? Do I need an S3 account for testing?

/claim #491

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -100,6 +100,9 @@ gem 'lograge'
 gem 'aws-sdk-s3', '~> 1'
 gem 'semverse'
 
+# Rubygems release engine
+gem 'compact_index'
+
 # Misc
 gem 'null_association'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -135,6 +135,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
+    compact_index (0.15.0)
     concurrent-ruby (1.3.1)
     connection_pool (2.4.1)
     crack (0.4.5)
@@ -530,6 +531,7 @@ DEPENDENCIES
   bcrypt (~> 3.1.7)
   bullet (~> 7.1.6)
   byebug
+  compact_index
   cucumber-rails (~> 2.5)
   cuke_modeler (~> 3.19)
   database_cleaner (~> 2.0)
--- a/app/controllers/api/v1/release_engines/rubygems/specs_controller.rb
+++ b/app/controllers/api/v1/release_engines/rubygems/specs_controller.rb
@@ -1,0 +1,162 @@
+# frozen_string_literal: true
+
+require 'open-uri'
+require 'rubygems/package'
+require 'compact_index'
+
+module Api::V1::ReleaseEngines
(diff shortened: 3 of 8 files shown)


ezekg (changes requested):
Thanks for this initial POC implementation. I'd prefer to see a service-based approach, rather than lumping it all into the controller, since the complexity here is higher than I originally expected it to be (not your fault, just the reality of how Rubygems functions). A service-based approach will allow us to write unit tests for everything. I'm cool with keeping it under the `Keygen` lib namespace, e.g. `Keygen::Packages::Engines::Rubygems`.

As the initial review mentions, I'd also like to explore storing the required gemspec contents locally in the database to avoid an HTTP round-trip every request, which would negatively effect latency and overall throughput. For [npm](https://github.com/keygen-sh/keygen-api/issues/489), I was thinking of throwing the entire `package.json` into the release's `metadata`, but having a new table for that e.g. `release_metadata`, may make more sense, and allow us to use that table for future package managers as well, like here with Rubygems.

I'm still trying to envision the overall API flow there, though.

What we _could_ do is kick off a job after upload that does download the gemspec contents from S3, and then saves it (or a `jsonb` representation of it) locally. The [artifact upload job](https://github.com/keygen-sh/keygen-api/blob/master/app/workers/wait_for_artifact_upload_worker.rb) might be a good place to kick off specialized 'processing' and storage of release 'metadata' like this, e.g. Rubygem gemspec and npm `package.json`, before marking the artifact as "ready."

E.g. if the artifact job is for a gemspec, before marking the artifact as "uploaded", we kick off an additional job to 'process' its gemspec by storing it in `release_metadata` table. We could even pre-generate some of these required gem index parts (e.g. `CompactIndex::GemVersion`), so that we aren't superfluously generating indexes each time a gem is downloaded.

Re: authentication. You'll likely need to either set the product to use an open distribution strategy (i.e. for no auth), or use [license key authentication](https://keygen.sh/docs/api/authentication/#license-authentication).

Re: backend. You'll need to create a test Cloudflare R2 or AWS S3 bucket to fully test things. There's no local filesystem backend at the moment.

Let me know if you're still feeling up for this? I'll update the main issue soon to mention this new direction so everything's clear.

wdyt?

nashby:
@ezekg hey, thanks for the review and sorry it took so long to respond. Looks like I won't have time to finish it properly so I'm going to close it for now. Feel free to reopen if you think it could be used as a reference for someone who's going to work on it. Sorry!

PR #907, opened 2024-10-18
Add support for Rubygems engine

### LINKED ISSUE #491: Add a release engine for Rubygems

ezekg:
Closes #491. In the end, it should be as simple as `gem build example.gemspec` and `keygen upload example-1.0.0.gem` — the rest should be automatically done by the engine.

## Prereqs

- [x] Add SHA256 checksum flag to CLI: https://github.com/keygen-sh/keygen-cli/issues/19
- [x] Add integration tests for compact index (e.g. `/versions`, `info/*`, etc.).
- [x] Add integration tests for specs (e.g. `/quick/*`, `/specs.4.8.gz`, etc.).
- [x] Add integration tests for gems (e.g. `/gems/*`).
- [x] Add unit tests for gem workers.

## Subreqs

- [x] Set up `rubygems.pkg.keygen.sh` subdomain.
- [ ] Update dashboard.
- [ ] Update changelog.
- [ ] Update docs.
- [ ] Reach out to leads/customers who asked for a gem engine?


ezekg:
Helpful: https://andre.arko.net/2014/03/28/the-new-rubygems-index-format/

ezekg:
Also helpful: https://github.com/rubygems/guides/blob/main/rubygems-org-compact-index-api.md

app/controllers/api/v1/release_engines/rubygems/specs_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    before_action :set_artifact
+
+    def quick
+      authorize! artifact,
+        to: :show?
+
+      # rubygems expects a marshalled and gzipped gemspec
+      gemspec = artifact.specification.as_gemspec
+      dump    = Marshal.dump(gemspec)
+      gz      = Zlib::Deflate.deflate(dump)
```
ezekg, line 17: `gz` is 'incorrect', maybe use `rz` or `compressed` — or create a `send_compressed_data` method?

app/controllers/api/v1/release_engines/rubygems/specs_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    before_action :authenticate_with_token
+    before_action :set_artifact
+
+    def quick
+      authorize! artifact,
+        to: :show?
+
+      # rubygems expects a marshalled and gzipped gemspec
+      gemspec = artifact.specification.as_gemspec
+      dump    = Marshal.dump(gemspec)
```
ezekg, line 16: Use a name like `serialized` instead.

app/workers/process_gem_specification_worker.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'rubygems/package'
+
+class ProcessGemSpecificationWorker < BaseWorker
```
ezekg, line 5: Add unit tests for this.

app/workers/wait_for_artifact_upload_worker.rb
```diff
@@ -26,22 +26,36 @@ def perform(artifact_id, enqueued_at = Time.current.iso8601)
...
+    )
+
+    NotifyArtifactUploadWorker.perform_async(
+      artifact.id,
+      artifact.status,
     )
+
+    # check if it's a specification e.g. gem package, etc.
+    case artifact
+    in filetype: ReleaseFiletype(:gem), engine: ReleaseEngine(:rubygems) if artifact.content_length.in?(ReleaseSpecification::MIN_CONTENT_LENGTH..ReleaseSpecification::MAX_CONTENT_LENGTH)
```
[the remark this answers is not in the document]
ezekg: I say it's expected behavior. New releases and artifacts will be able to take advantage of the engine.

app/models/release.rb
```diff
@@ -250,16 +258,31 @@ class Release < ApplicationRecord
...
-    reorder(<<~SQL.squish)
-      releases.semver_major      DESC,
-      releases.semver_minor      DESC NULLS LAST,
-      releases.semver_patch      DESC NULLS LAST,
-      releases.semver_pre_word   DESC NULLS FIRST,
-      releases.semver_pre_num    DESC NULLS LAST,
-      releases.semver_build_word DESC NULLS LAST,
-      releases.semver_build_num  DESC NULLS LAST
-    SQL
+  scope :order_by_version, -> (order = :desc) {
```
ezekg, line 261: Add unit tests for the new sort order arg.

spec/factories/release_specification.rb
```diff
@@ -0,0 +1,18 @@
...
+
+    account     { NIL_ACCOUNT }
+    environment { NIL_ENVIRONMENT }
+    artifact    { build(:artifact, account:, environment:) }
+    release     { artifact.release }
+    content     { SecureRandom.bytes(128) }
+
+    trait :rubygems do
+      artifact { build(:artifact, :rubygems, account:, environment:) }
+      content  { file_fixture('ping-1.0.0.gem').read }
```
ezekg, line 15: Should be a gemspec not a gem.

app/models/release_artifact.rb
```diff
@@ -267,7 +289,7 @@ class ReleaseArtifact < ApplicationRecord
                .uploaded,
          )
          .reorder(
-           created_at: DEFAULT_SORT_ORDER,
+           "#{table_name}.created_at": DEFAULT_SORT_ORDER,
```
ezekg, line 292: why?

app/models/release_artifact.rb
```diff
@@ -398,24 +430,28 @@ class ReleaseArtifact < ApplicationRecord
 
     case
     when license.revoke_access?
-      license.expired? ? none : joins(:release).where(release: { created_at: ..license.expiry })
+      license.expired? ? none : joins(:release).where(releases: { created_at: ..license.expiry })
     when license.restrict_access?,
          license.maintain_access?
-      joins(:release).where(release: { created_at: ..license.expiry })
+      joins(:release).where(releases: { created_at: ..license.expiry })
```
ezekg, line 436: why???

app/workers/wait_for_artifact_upload_worker.rb
```diff
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 class WaitForArtifactUploadWorker < BaseWorker
+  MIN_CONTENT_LENGTH = 5.bytes      # to avoid processing empty or invalid artifacts
+  MAX_CONTENT_LENGTH = 25.megabytes # to avoid downloading large artifacts
```
ezekg, line 5: Is this large enough if we look at popular gems?

ezekg: Looking at popular gems, most are well under 1mb.

app/models/release_artifact.rb
ezekg: Reason: unions don't always know what table they're for, since a union could be across tables.

app/models/release_artifact.rb
ezekg: Rails needs to be smarter here.

app/controllers/api/v1/release_engines/rubygems/compact_index_controller.rb
```diff
@@ -0,0 +1,116 @@
...
+      authorize! artifacts,
+        to: :index?
+
+      gems = artifacts.group_by(&:package).reduce([]) do |arr, (package, artifacts)|
+        versions = to_versions(artifacts)
+
+        arr << CompactIndex::Gem.new(package.key, versions.sort)
+      end
+
+      render plain: CompactIndex.versions(
```
ezekg: Should we add etag support? Looks like gem internals use it for caching purposes to reduce load.

ezekg:
After some testing, `bundle install` works fine, but `gem install` emits a warning and has issues resolving dependencies if they're internal because it relies on the `specs.4.8.gz` file, which we don't currently support. So much for only implementing the compact index… looks like the legacy spec files are also required. 🤦‍♂️

ezekg:
Relevant: https://github.com/rubygems/rubygems/issues/8012#issuecomment-2372235771

ezekg:
The dependency API that I've seen [others](https://github.com/gemfast/server/blob/f42c0ba7dfef0ad49812b1a6ab6fdd5d864287a1/internal/api/api.go#L199) implement is no longer needed, at least for Bundler: https://blog.rubygems.org/2023/02/22/dependency-api-deprecation

app/controllers/concerns/compression.rb
```diff
@@ -0,0 +1,18 @@
...
+module Compression
+  extend ActiveSupport::Concern
+
+  def deflate(data, **) = Zlib::Deflate.deflate(data, **)
+  def gzip(data, deterministic: true, **)
+    zipped = StringIO.new
+    zipped.set_encoding(Encoding::BINARY)
+
+    gz = Zlib::GzipWriter.new(zipped, Zlib::BEST_COMPRESSION)
+    gz.mtime = 0 if deterministic
```
ezekg, line 12: Gzip has an internal timestamp by default, which makes it non-deterministic, which breaks etags.

app/models/application_record.rb
```diff
@@ -19,7 +19,8 @@ class ApplicationRecord < ActiveRecord::Base
     prepend: true
 
   default_scope -> {
-    order(created_at: DEFAULT_SORT_ORDER)
+    # FIXME(ezekg) why do we need an explciit table_name everywhere?
+    order("#{table_name}.created_at": DEFAULT_SORT_ORDER)
```
ezekg, line 23: Simple answer: unions. They some unions don't have a table, since they potentially union across multiple tables (theoretically — but we always union the same table, iirc). This — arguably ugly — explicitness makes sure that we don't hit ambiguous column errors with a bare `created_at` ordering.

PR #914, opened 2024-11-08
Add support for OCI engine

### LINKED ISSUE #911: Add support for Docker/OCI container registry
What if we did [an OCI-compliant registry](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints) for Docker? Can test [conformance here](https://github.com/opencontainers/distribution-spec/tree/main/conformance).

_Originally posted by @ezekg in https://github.com/keygen-sh/keygen-api/issues/642#issuecomment-1184513875_

ezekg:
Closes #911.  In the end, it should be as simple as building an OCI image layout via `docker buildx build --platform linux/amd64,linux/arm64 -o type=oci,dest=- . > keygen-1.0.0.tar` and `keygen upload keygen-1.0.0.tar` — the rest should be automatically handled by the OCI engine. Afterwards, `docker pull oci.pkg.keygen.sh/keygen/api` should work.

Like with npm and friends, `docker push` will not be supported right away.

## Prereqs

- [x] Implement the [Docker/OCI manifest and blob API endpoints](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests).
- [x] Test [parsing](https://github.com/keygen-sh/keygen-api/blob/da5fa6b6256c1c7cf6189464f7faa3354e7f4c3a/lib/oci_image_layout.rb) of internal/popular images (e.g. Alpine, Ubuntu, Keygen, Relay, etc.)
- [x] Test multi-platform images using `docker buildx`.
- [x] Test [OCI conformance](https://github.com/opencontainers/oci-conformance) (namely `OCI_TEST_PULL=1`).[^0]
- [x] Write tests for OCI spec conformance.
- [x] Write tests for manifests API.
- [x] Write tests for blobs API.
- [x] Write tests for tags API.
- [ ] ~~Helm compatibility (should be fine?)~~ Can [dig into this](https://github.com/helm/helm/issues/13296#issuecomment-2605851839) later.
- [x] Make EE-only.
- [x] Cut a self-hosted release beforehand.

## Subreqs

- [x] Set up `oci.pkg.keygen.sh` subdomain.
- [ ] Update dashboard.
- [ ] Update changelog.
- [ ] Update website (e.g. landing pages, EE features, etc.)
- [ ] Update blog.
- [ ] Update docs. (Note: add new webhook events!)
- [ ] Outreach.
- [ ] Stdout?
- [ ] [Apply?](https://conformance.opencontainers.org)

[^0]: Results from the OCI conformance test can be found [here](https://gist.github.com/ezekg/00198354b80f9f931a2b7dab81e7f7fd).


spec/workers/process_docker_image_worker_spec.rb
```diff
@@ -0,0 +1,321 @@
...
+
+      it 'should not process image' do
+        expect { subject.perform_async(artifact.id) }.to not_change { artifact.reload.manifest }
+
+        expect(artifact.status).to eq 'FAILED'
+      end
+    end
+  end
+
+  context 'when artifact filesize is unaccurate' do
```
ezekg, line 295: unaccurate? lol

spec/support/matchers/upload_matcher.rb
```diff
@@ -0,0 +1,44 @@
...
+    )
+
+    block.call
+
+    # FIXME(ezekg) should this compare sizes >=?
+    expectations.empty? ? !uploads.empty? : matches.size == expectations.size
+  end
+
+  failure_message do
+    "Expected block to upload files matching #{expected.inspect}, but it wasn't uploaded: #{uploads.inspect}"
```
ezekg, line 38: s/files/objects/

app/workers/process_docker_image_worker.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'minitar'
+require 'zlib'
+
+class ProcessDockerImageWorker < BaseWorker
+  MIN_TARBALL_SIZE  = 5.bytes    # to avoid processing empty or invalid tarballs
```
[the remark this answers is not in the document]
ezekg: > The minimum filesize for a valid uncompressed tarball is 512 bytes. This comes from the minimum block size required by the tar format, which is 512 bytes.
>
> A completely empty tarball (no files or metadata) will have at least two 512-byte blocks: one for the tar header and one for the end-of-archive marker, totaling 1024 bytes. So, in practice, the smallest valid empty tar file is usually 1024 bytes.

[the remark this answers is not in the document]
ezekg: > For a gzipped tarball (a .tar.gz file), the minimum size is generally around 28 bytes. This is the smallest valid gzip header, footer, and minimal compressed data needed to indicate an empty file.
>
> If the tarball itself is empty (like a 1024-byte uncompressed tarball containing no files), gzip will compress it down significantly. The exact size depends on the gzip implementation and options but is typically in the range of 28 to 40 bytes.

app/workers/process_docker_image_worker.rb
```diff
@@ -0,0 +1,118 @@
+# frozen_string_literal: true
+
+require 'minitar'
+require 'zlib'
+
+class ProcessDockerImageWorker < BaseWorker
```
ezekg, line 6: This worker should be in a `high_memory` queue.

ezekg:
Helpful: https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pulling-manifests

app/workers/process_docker_image_worker.rb
```diff
@@ -0,0 +1,118 @@
...
+
+    # keep a ref to the manifest
+    manifest = nil
+
+    # unpack the package tarball
+    unpack tar do |archive|
+      archive.each do |entry|
+        case entry.name
+        in 'index.json'
+          # TODO(ezekg) do we need this for OCI compliance or can we generate dynamically?
```
ezekg, line 42: This is required for multi-platform images. Clients can request it using the `application/vnd.docker.distribution.manifest.list.v2+json` content type, vs the `application/vnd.docker.distribution.manifest.v2+json` content type.

ezekg: I think we can actually use this in place of `manifest.json`, since  it's Docker-specific and seems `index.json` replaces it, at least for the OCI.

app/workers/process_docker_image_worker.rb
```diff
@@ -0,0 +1,118 @@
...
+          )
+        in %r{^blobs/sha256/} if entry.file?
+          key = artifact.key_for(entry.name)
+
+          # skip if already uploaded
+          next if
+            client.head_object(bucket: artifact.bucket, key:).successful? rescue false
+
+          # upload blob in chunks
+          client.put_object(bucket: artifact.bucket, key:) do |writer|
```
ezekg, line 69: Set `content_type` to appropriate blob media type per the manifest.

app/controllers/api/v1/release_engines/oci/blobs_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    private
+
+    attr_reader :package
+
+    def set_package
+      scoped_packages = authorized_scope(current_account.release_packages.oci)
+                          .where_assoc_exists(
+                            %i[releases artifacts manifest], # must exist
```
ezekg, line 27: This should check for descriptors not manifests.

app/controllers/api/v1/release_engines/oci/manifests_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    end
+
+    private
+
+    attr_reader :package
+
+    def set_package
+      scoped_packages = authorized_scope(current_account.release_packages.oci)
+                          .where_assoc_exists(
+                            %i[releases artifacts manifest], # must exist
```
ezekg, line 33: This could be simplified to manifests.

app/controllers/application_controller.rb
```diff
@@ -86,6 +86,8 @@ def render_forbidden(**kwargs)
 
     respond_to do |format|
       format.any {
+        response.content_type = Mime::Type.lookup_by_extension(:jsonapi)
```
ezekg, line 89: I think we should respond with a JSON content type if asked (or if the route uses the `:json` format), instead of always responding with JSONAPI.

app/controllers/concerns/rendering.rb
```diff
@@ -8,17 +8,24 @@ module Base
...
+        # skip if we've already set content type
+        unless mime_type.nil?
+          return super
+        end
+
+        case options
+        in jsonapi: _
           response.content_type = Mime::Type.lookup_by_extension(:jsonapi)
-        in json:
+        in json: _
```
ezekg, line 22: This can respond with JSONAPI depending on route format and accept header, since JSON and JSONAPI are often synonymous.

app/workers/wait_for_artifact_upload_worker.rb
```diff
@@ -54,6 +54,14 @@ def perform(artifact_id, enqueued_at = Time.current.iso8601)
       )
 
       ProcessNpmPackageWorker.perform_async(artifact.id)
+    in filetype: ReleaseFiletype(:tar), engine: ReleaseEngine(:oci)
```
ezekg, line 57: Support other filetypes?

db/migrate/20241114163421_add_content_attributes_to_release_manifests.rb
```diff
@@ -0,0 +1,14 @@
+class AddContentAttributesToReleaseManifests < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :release_manifests, :content_path,   :string, null: true
+    add_column :release_manifests, :content_type,   :string, null: true
+    add_column :release_manifests, :content_length, :bigint, null: true
+    add_column :release_manifests, :content_digest, :string, null: true
+
+    add_index :release_manifests, %i[release_artifact_id content_digest],
```
ezekg, line 10: Content path should also be unique ala descriptors

features/api/v1/engines/tauri/v2/upgrade.feature
```diff
@@ -84,7 +84,7 @@ Feature: Tauri v2 upgrade application
     Then the response status should be "403"
     And the response should contain the following headers:
       """
-      { "Content-Type": "application/json; charset=utf-8" }
+      { "Content-Type": "application/vnd.api+json; charset=utf-8" }
```
ezekg, line 87: Should this respond in JSON since the route's format is JSON?

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,279 @@
...
+module OciImageLayout
+  class Error < StandardError; end
+  class LayoutNotFoundError < Error; end
+  class IndexNotFoundError < Error; end
+  class UnsupportedMediaTypeError < Error; end
+  class BadDigestError < Error; end
+
+  def self.parse(io) = Parser.new(io).parse
+
+  class DigestIO < SimpleDelegator
```
ezekg, line 16: Move to lib

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,279 @@
...
+        Manifest.from_json(manifest, fs:)
+      end
+
+      io.rewind
+    end
+
+    def each(&)
+      super
+
+      manifests.each { _1.each(&) }
```
ezekg, line 205: Why can't we just do `manifests.each(&)`?

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,279 @@
...
+    def initialize(io, fs:, **)
+      super(io, fs:, **)
+
+      # FIXME(ezekg) add support for annotations? e.g. auto-tagging
+      @annotations = []
+      @manifests   = []
+      @layers      = []
+      @config      = nil
+
+      # some blobs may not be referenced without being present
```
ezekg, line 247: s/may not be/may be/

spec/factories/release_descriptor.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+require 'rubygems/package'
```
ezekg, line 3: Nix requires

spec/factories/release_manifest.rb
```diff
@@ -12,11 +12,16 @@
     environment { NIL_ENVIRONMENT }
     artifact    { build(:artifact, account:, environment:) }
     release     { artifact.release }
-    content     { Random.bytes(128) }
+
+    content_type   { 'application/octet-stream' }
```
ezekg, line 16: Add content path

spec/factories/release_manifest.rb
```diff
@@ -40,5 +47,23 @@
...
+      artifact       { build(:artifact, :oci_image, account:, environment:) }
+      content_digest { "sha256:#{Digest::SHA256.hexdigest(content)}" }
+      content_type   { 'application/vnd.oci.image.index.v1+json' }
+      content        {
+        tar   = file_fixture('alpine-3.20.3.tar').read
+        index = Minitar::Reader.open tar do |archive|
+          json = archive.find { _1.name in 'index.json' }
+                        .read
+
+          JSON.parse(json) # minify
```
ezekg, line 61: Skip

spec/support/matchers/upload_matcher.rb
```diff
@@ -0,0 +1,51 @@
...
+  objects = []
+  matches = []
+
+  match do |block|
+    allow_any_instance_of(Aws::S3::Client).to(
+      receive(:put_object).and_wrap_original do |original, **object, &writer|
+        objects << object
+
+        expectations.each do |expected|
+
```
ezekg, line 17: Nix newline

spec/workers/process_oci_image_worker_spec.rb
```diff
@@ -0,0 +1,355 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require 'spec_helper'
+
+require 'minitar'
+require 'zlib'
```
ezekg, line 7: Nix zlib

spec/workers/process_oci_image_worker_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+      it 'should change artifact status' do
+        expect { subject.perform_async(artifact.id) }.to change { artifact.reload.status }.to('UPLOADED')
+      end
+
+      it 'should store manifests' do
+        expect { subject.perform_async(artifact.id) }.to change { artifact.reload.manifests }
+
+        expect(artifact.manifests).to satisfy { |manifests|
+          manifests in [
+            ReleaseManifest(content_path: 'index.json',                                                                    content_digest: 'sha256:355eee6af939abf5ba465c9be69c3b725f8d3f19516ca9644cf2a4fb112fd83b', content_type: 'application/vnd.oci.image.index.v1+json',                   content_length: 441,  content: String),
```
ezekg, line 53: Add content asserts

spec/workers/process_oci_image_worker_spec.rb
```diff
@@ -0,0 +1,355 @@
...
+    end
+
+    it 'should not process file' do
+      expect { subject.perform_async(artifact.id) }.to not_change { artifact.reload.manifest }
+
+      expect(artifact.status).to eq 'FAILED'
+    end
+
+    it 'should be efficient' do
+      expect { subject.perform_async(artifact.id) }.to allocate_less_than(5.megabytes)
```
ezekg, line 352: Refactor this to sample memory usage and assert against peak

app/controllers/api/v1/release_artifacts_controller.rb
```diff
@@ -39,7 +39,10 @@ def show
       return render jsonapi: artifact if
         !artifact.downloadable? || prefers?('no-download')
 
-      download = artifact.download!(ttl: release_artifact_query[:ttl])
+      download = artifact.download!(
+        filename: params.fetch(:filename) { artifact.filename },
```
ezekg, line 43: Add tests.

lib/enumerator_io.rb
```diff
@@ -0,0 +1,37 @@
...
+    # extract the requested amount of data from the buffer
+    if length
+      buffer << @buffer.slice!(0, length)
+    else
+      buffer << @buffer
+
+      @buffer.clear
+    end
+
+    buffer.empty? ? nil : buffer
```
ezekg, line 35: `buffer.presence` should suffice?

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,280 @@
...
+module OciImageLayout
+  class Error < StandardError; end
+  class LayoutNotFoundError < Error; end
+  class IndexNotFoundError < Error; end
+  class UnsupportedMediaTypeError < Error; end
+  class BadDigestError < Error; end
+
+  def self.parse(io) = Parser.new(io).parse
+
+  class DigestIO < SimpleDelegator
```
ezekg, line 17: Move to separate lib

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,280 @@
...
+      super(io, media_type: 'application/vnd.oci.layout.header.v1+json', **)
+
+      data     = JSON.parse(io.read)
+      @version = data['imageLayoutVersion']
+
+      io.rewind
+    end
+  end
+
+  class Index < Blob
```
ezekg, line 186: Store manifest version?

app/workers/wait_for_artifact_upload_worker.rb
ezekg: For example `tar.gz` and `tgz`.

app/models/release_manifest.rb
```diff
@@ -47,6 +48,225 @@ class ReleaseManifest < ApplicationRecord
...
+    base = joins(:release)
+
+    manifests = base.where(content_digest: reference)
+                    .or(base.where(releases: { version: reference }))
+                    .or(base.where(releases: { tag: reference }))
+
+    # oci clients may want a specific media type in case of multiple manifests
+    unless content_type.blank?
+      manifests = case content_type
+                  in ['*/*', *] | '*/*' # has no preference i.e. accepts anything
```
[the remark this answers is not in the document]
ezekg: It's not.

db/migrate/20241114163421_add_content_attributes_to_release_manifests.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    add_index :release_manifests, %i[content_path release_artifact_id],
+      algorithm: :concurrently,
+      unique: true
+
+    add_index :release_manifests, %i[content_digest release_artifact_id],
+      algorithm: :concurrently,
+      unique: true
+
+    add_index :release_manifests, %i[content_type release_artifact_id],
```
ezekg, line 18: Doesn't need artifact in the index.

db/migrate/20241113134529_create_release_descriptors.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      t.timestamps
+
+      t.index %i[account_id created_at],              order: { created_at: :desc }
+      t.index %i[environment_id]
+      t.index %i[release_id]
+      t.index %i[release_artifact_id]
+      t.index %i[content_path release_artifact_id],   unique: true
+      t.index %i[content_digest release_artifact_id], unique: true
+      t.index %i[content_type release_artifact_id]
```
ezekg, line 23: Drop artifact from index.

db/migrate/20241114163421_add_content_attributes_to_release_manifests.rb
```diff
@@ -0,0 +1,21 @@
+class AddContentAttributesToReleaseManifests < ActiveRecord::Migration[7.2]
+  disable_ddl_transaction!
+
+  def change
+    add_column :release_manifests, :content_path,   :string, null: true
```
ezekg, line 5: Eventually make these not-null.

app/workers/process_oci_image_worker.rb
```diff
@@ -0,0 +1,135 @@
+# frozen_string_literal: true
+
+require 'minitar'
+
+require Rails.root / 'lib/oci_image_layout'
+
+class ProcessOciImageWorker < BaseWorker
+  MIN_TARBALL_SIZE  = 512.bytes     # to avoid processing empty or invalid tarballs
+  MAX_TARBALL_SIZE  = 512.megabytes # to avoid downloading excessive tarballs
```
ezekg, line 9: This is unlikely to be sufficient for some applications, but I feel like it's a good default. Maybe bumping it up to 1GB would be fine, since I've seen quite a few 800MB images. I think we should eventually add a max artifact size to accounts.

ezekg: E.g. Sentry is 250MB and PostHog is 1.5GB.

ezekg: Maybe add `max_upload` for individual uploads, and `max_ingress` and `max_egress` for monthly ingress/egress limits (which can be handled like `max_reqs`).

features/api/v1/engines/oci/manifests.feature
```diff
@@ -0,0 +1,1580 @@
...
...                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                         | 2024-11-11T10:26:01.000Z | 2024-11-11T10:26:01.000Z |
```
ezekg, line 52: Need to drop `index.json` manifests since they aren't meant to be served by a registry (oops).

(Fixed in 50bb888841cca62eac476152ba3d743490c2af0b.)

app/models/release_manifest.rb
```diff
@@ -234,7 +234,7 @@ def as_package_json = JSON.parse(content)
   # which could be a digest, version, or tag, and also to request a specific media
   # type in case of multiple manifests per-image.
   def self.find_by_reference!(reference, content_type: nil)
-    base = joins(:release)
+    base = reorder(created_at: :asc).joins(:release)
```
ezekg, line 237: Revert?

ezekg: Apparently, Docker is retarded and it doesn't actually do any kind of meaningful content negotiation. Instead, it just asks for anything, `application/vnd.docker.*` and `application/vnd.oci.*`, and in a random order too.

So this needs to be reworked, and how we store and retrieve manifests needs to be more sophisticated, because unless `reference` is a digest, we always want to retrieve the root manifest (i.e. the index for a multi-platform image).

app/models/release_manifest.rb
```diff
@@ -47,6 +48,225 @@ class ReleaseManifest < ApplicationRecord
...
+  # which could be a digest, version, or tag, and also to request a specific media
+  # type in case of multiple manifests per-image.
+  def self.find_by_reference!(reference, content_type: nil)
+    base = reorder(created_at: :asc).joins(:release)
+
+    manifests = base.where(content_digest: reference)
+                    .or(base.where(releases: { version: reference }))
+                    .or(base.where(releases: { tag: reference }))
+
+    # oci clients may want a specific media type in case of multiple manifests
```
ezekg, line 243: So that was a lie.

app/models/release_artifact.rb
```diff
@@ -38,11 +38,22 @@ class ReleaseArtifact < ApplicationRecord
     inverse_of: :artifacts,
     autosave: true,
     optional: true
+  has_many :manifests,
+    class_name: 'ReleaseManifest',
+    foreign_key: :release_artifact_id,
+    inverse_of: :artifact,
+    dependent: :delete_all
+  # NOTE(ezekg) not a strict has-one but this is a convenience
```
ezekg, line 46: Should we revert back to has-one???

app/controllers/api/v1/release_engines/oci/manifests_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    before_action :scope_to_current_account!
+    before_action :require_active_subscription!
+    before_action :authenticate_with_token
+    before_action :set_package
+
+    def show
+      authorize! package
+
+      manifest = authorized_scope(package.manifests).find_by_reference!(
+        params[:reference],
```
ezekg, line 15: If this is a digest, we should redirect to blobs.

app/workers/process_oci_image_worker.rb
```diff
@@ -0,0 +1,140 @@
...
+    # unpack the image tarball
+    unpack tar do |archive|
+      layout = OciImageLayout.parse(archive)
+
+      layout.each do |item|
+        next unless
+          item.exists?
+
+        # store manifests as a manifest for easy reference without hitting the storage provider
+        if item in OciImageLayout::Manifest => manifest
```
ezekg, line 47: There should only be one manifest.

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,255 @@
...
+      # some blobs may be referenced without being present
+      return unless
+        exists?
+
+      data = JSON.parse(io.read)
+      io.rewind
+
+      if manifests = data['manifests']
+        manifests.each do |manifest|
+          @manifests << Manifest.from_json(manifest, fs:)
```
ezekg, line 232: This should be a `Descriptor` per the spec, afaict.

app/controllers/api/v1/release_engines/oci/tags_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+    before_action :authenticate_with_token
+    before_action :set_package
+
+    def index
+      authorize! package, to: :show?
+
+      releases = authorized_scope(package.releases)
+      authorize! releases
+
+      tags = releases.pluck(:tag)
```
ezekg, line 17: Needs to be in lexical order (i.e. case-insensitive alphanumeric order) per the spec.

app/workers/process_oci_image_worker.rb
ezekg: This was wrong.

app/controllers/api/v1/release_engines/oci/manifests_controller.rb
ezekg: Above assumption was wrong.

app/models/release_artifact.rb
ezekg: No.

lib/oci_image_layout.rb
```diff
@@ -0,0 +1,324 @@
...
+      super
+
+      layers.each { _1.each(&) }
+      config&.each(&)
+    end
+  end
+
+  class Config < Descriptor
+    def self.from_descriptor(descriptor, **)
+      raise "expected config descriptor but got #{descriptor.media_type}" unless
```
ezekg, line 309: This should not raise on unknown config media types.

PR #926, opened 2025-01-23
Fix Docker authentication

ezekg:
Surprises never stop coming. This refactors the OCI engine to follow Docker's authentication dance per the [distribution registry spec](https://github.com/distribution/distribution/blob/main/docs/content/spec/auth/token.md). Essentially, even if registry credentials are added via `docker login oci.pkg.keygen.sh`, Docker won't send the credentials unless the registry first sends a 401 Unauthorized status code (imo a waste of time and bandwidth but wtfe).

The downside to this is that we now leak information by responding with 401/403s instead of 404s, but since this is limited only to packages using an OCI engine, I think the risk is low and it can be considered expected behavior.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -107,7 +107,16 @@ class ApplicationController < ActionController::API
   def render_unauthorized(**kwargs)
     skip_verify_authorized!
 
-    self.headers['WWW-Authenticate'] = %(Bearer realm="keygen")
+    # FIXME(ezekg) docker wants to do a jwt token dance unless we stick to a basic
+    #              auth scheme (which is fine since our basic auth scheme only
+    #              supports license keys and tokens, not passwords!)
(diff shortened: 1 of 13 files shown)


app/controllers/application_controller.rb
```diff
@@ -107,7 +107,16 @@ def render_forbidden(**kwargs)
...
+    # FIXME(ezekg) docker wants to do a jwt token dance unless we stick to a basic
+    #              auth scheme (which is fine since our basic auth scheme only
+    #              supports license keys and tokens, not passwords!)
+    default_challenge_scheme = oci? ? 'Basic' : 'Bearer'
+
+    challenge_scheme = authentication_scheme&.capitalize || default_challenge_scheme
+    challenge_realm  = 'keygen'
+    challenge        = %(#{challenge_scheme} realm="#{challenge_realm}")
+
+    response.headers['WWW-Authenticate'] = challenge
```
ezekg, line 119: Should probably refactor this out into the authentication concern.

PR #935, opened 2025-02-24
Add support for session authentication

ezekg:
Groundwork for #409 and Portal.


app/controllers/api/v1/tokens_controller.rb
```diff
@@ -82,7 +82,16 @@ def generate
...
 
+      # TODO(ezekg) make default session expiry configurable
+      session = token.sessions.build(
+        expiry: token.expiry.presence || 1.week.from_now,
+        user_agent: request.user_agent,
+        ip: request.remote_ip,
+      )
+
       if token.save
+        cookies.encrypted[:session_id] = { value: session.id, httponly: true, secure: true, same_site: :none, expires: session.expiry, domain: Keygen::DOMAIN }
```
ezekg, line 93: What if we're generating a token on behalf of another bearer, but don't want to clobber our session? E.g. an admin generating an environment token for an integration. I guess use the object's tokens relationship instead of the root? E.g. `POST /v1/environments/<environment>/tokens` instead of `/v1/tokens`?

ezekg: Or we could start a session only on `:password` authn?

app/models/session.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+class Session < ApplicationRecord
+  MAX_AGE = 2.weeks
+
+  include Denormalizable
+  include Environmental
+  include Accountable
+
+  belongs_to :token
```
ezekg, line 10: How will this work with SSO, i.e. not tied to a token? Maybe the token association should be optional?

ezekg: But then we'd need to expose sessions so that they can be revoked.

May be easier to simply generate a token as part of the SSO flow.

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -139,6 +154,11 @@ def revoke
         resource: token,
       )
 
+      # expire session if we're revoking the current token
+      if token == current_token
+        cookies.delete(:session_id, domain: Keygen::DOMAIN, same_site: :none)
```
ezekg, line 159: Is all this really needed? Invalid sessions will be auto-expired after unsuccessful the authn.

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -127,6 +137,11 @@ def regenerate
         resource: token,
       )
 
+      # expire session if we're regenerating the current token
+      if token == current_token
+        cookies.delete(:session_id, domain: Keygen::DOMAIN, same_site: :none)
```
ezekg, line 142: Do we _want_ to expire the session when a token is regenerated?

features/api/v1/tokens/sessions.feature
```diff
@@ -239,9 +239,33 @@ Feature: Token sessions
     And I authenticate with a session
     When I send a PUT request to "/accounts/test1/tokens/$2"
     Then the response status should be "200"
-    And the response headers should not contain "Set-Cookie"
```
ezekg: Put back

app/controllers/api/v1/tokens_controller.rb
ezekg: Or should we revoke all sessions and generate a new session entirely? I lean towards that.

ezekg: `regenerate!` can return the new session.

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -112,36 +112,33 @@ def regenerate_current
       authorize! current_token,
         to: :regenerate?
 
-      current_token.regenerate!
+      if session = current_token.regenerate!(session: current_session)
+        cookies.encrypted[:session_id] = { value: session.id, httponly: true, secure: true, same_site: :none, expires: session.expiry, domain: Keygen::DOMAIN }
```
ezekg, line 116: Refactor this out into a method?

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -154,9 +151,9 @@ def revoke
         resource: token,
       )
 
-      # expire session if we're revoking the current token
-      if token == current_token
-        cookies.delete(:session_id, domain: Keygen::DOMAIN, same_site: :none)
+      # expire current session if we're revoking its token
+      unless current_session.nil?
+        cookies.delete(:session_id, domain: Keygen::DOMAIN, same_site: :none) if current_session.token == token
```
ezekg, line 156: Refactor this out into a method?

app/controllers/api/v1/tokens_controller.rb
ezekg: I say we leave as-is so that impersonation is possible. Cookies/credentials can always be ignored if this isn't desired, or the requestor can use one of the other token relationships instead.

app/models/session.rb
ezekg: Although, exposing sessions would, in the long run, probably be a good idea from a security perspective. We could show a page on the user's dashboard listing active tokens and sessions, letting them revoke as-needed.

ezekg: Fine with doing that in a follow up PR, though, since it's not directly related to session authn.

lib/keygen/middleware.rb
```diff
@@ -283,5 +283,21 @@ def call(env)
...
+    # NOTE(ezekg) see: https://github.com/rack/rack/issues/2130
+    class PartitionedCookies
+      def initialize(app) = @app = app
+      def call(env)
+        status, headers, body = @app.call(env)
+
+        # add support for partitioned cookies
+        cookie = headers['Set-Cookie']
+        if cookie in /SameSite=None/
+          cookie.gsub!('SameSite=None', 'SameSite=None; partitioned;')
```
ezekg, line 296: See: https://github.com/mozilla/standards-positions/issues/678

app/controllers/concerns/cookies.rb
```diff
@@ -0,0 +1,32 @@
...
+  def set_session_id_cookie(session)
+    return unless session in Session # lol nice
+    return unless
+      request.origin.nil? || request.origin.ends_with?(Keygen::DOMAIN)
+
+    cookies.encrypted[:session_id] = {
+      value: session.id,
+      expires: session.expiry,
+      domain: Keygen::DOMAIN,
+      same_site: :lax,
```
ezekg, line 17: I'm so confused about same-site cookies esp i.r.t. CSRF. Everything online is ambiguous at best, or wrong at worst.

Goal is to allow `fetch()` with cookies sent from `.keygen.sh` origins, but not from `evil.example`.

app/controllers/concerns/authentication.rb
```diff
@@ -328,10 +328,12 @@ def request_http_basic_authentication(realm = 'keygen', message = nil)
     raise Keygen::Error::UnauthorizedError.new(code: 'TOKEN_INVALID')
   end
 
-  # we only support cookie authn from same- or sub-origins (i.e. subdomains)
+  # NOTE(ezekg) we only support cookie authn from same-site origins (e.g. subdomains) to prevent CSRF
+  #
+  #             see: https://scotthelme.co.uk/csrf-is-dead/
   def has_cookie_credentials?
     (request.origin.nil? || request.origin.ends_with?(Keygen::DOMAIN)) &&
```
ezekg, line 335: This introduces a CRSF attack vector from subdomains out of our control, like `status.keygen.sh`. We should either explicitly whitelist the domains that can use cookie authn, or move these third-party subdomains onto a separate domain, e.g. `keygenstatus.com`. The former is the safest route.

config/initializers/rack_cors.rb
```diff
@@ -8,29 +8,45 @@
...
+    headers: :any,
+    methods: :any,
+    max_age:,
+    expose:,
+  }
+
+  allow do
+    origins { _1.ends_with?(Keygen::DOMAIN) }
+
+    resource '*', **options, credentials: true
```
ezekg, line 44: We could add [a `X-CSRF-Token` header](https://security.stackexchange.com/questions/154410/how-does-x-csrf-token-work) and use that as an additional means of [preventing CSRF](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/CSRF_prevention#implement_an_anti-csrf_token_along_with_samesitestrict).

app/controllers/concerns/authentication.rb
```diff
@@ -96,10 +110,53 @@ def query_token_authenticator(query_token)
...
+
+    case
+    when session.bearer.has_role?(:license)
+      raise Keygen::Error::ForbiddenError.new(code: 'SESSION_NOT_ALLOWED', detail: 'Session authentication is not allowed by policy') unless
+        session.bearer.supports_session_auth?
+    end
+
+    if session.last_used_at.nil? || session.last_used_at.before?(1.hour.ago)
+      session.update(
+        expiry: session.expiry + 1.hour, # extend expiry while in use until MAX_AGE
```
ezekg, line 140: This should only extend expiry when it expires within a certain duration e.g. 1 hour.

app/controllers/concerns/authentication.rb
```diff
@@ -271,6 +330,14 @@ def request_http_basic_authentication(realm = 'keygen', message = nil)
     raise Keygen::Error::UnauthorizedError.new(code: 'TOKEN_INVALID')
   end
 
+  # NOTE(ezekg) we only support cookie authn from portal origin to prevent CSRF
+  #
+  #             see: https://scotthelme.co.uk/csrf-is-dead/
+  def has_cookie_credentials?
+    (request.origin&.ends_with?(Keygen::Portal::HOST)) &&
```
ezekg, line 337: This allows e.g. `evil-safe.com` if portal host were `safe.com`. I think we'd be okay with asserting direct match since we currently don't allow multiple origins (and even then we could introduce `KEYGEN_PORTAL_HOSTS`).

config/initializers/rack_cors.rb
```diff
@@ -45,7 +45,7 @@
   }
 
   allow do
-    origins { _1.ends_with?(Keygen::DOMAIN) }
+    origins Keygen::Portal::ORIGIN
```
ezekg, line 48: Similar vulnerability that triggered this change: https://github.com/zeromicro/go-zero/security/advisories/GHSA-fgxv-gw55-r5fq

ezekg:
When authenticating with a session cookie, should certain destructive or otherwise sensitive actions, such as generating an API token or deleting a product, require re-authentication? I think this would be a wise addition to not only prevent XSS, but also to improve the UX in Portal by preventing expensive mistakes, which have occurred in the past.

PR #940, opened 2025-04-04
Add denormalized policy association to machines

ezekg:
With more and more accounts utilizing machine heartbeats, we're starting to see slow queries for [the cull crons](https://github.com/keygen-sh/keygen-api/blob/master/app/workers/cull_dead_machines_worker.rb) because we have to join across licenses in order to get a machine's heartbeat duration from its policy. I've tried really hard to optimize that query, but there doesn't really seem to be a good way to do that due to table size. This issue will continue to get worse since culling is run every 2 minutes. Rather than try to move `next_heartbeat_at` from a virtual attribute to a column — and keep that in sync across policy changes — I decided to simply denormalize the policy onto the machine.

## Subreqs

- [ ] Run `VACUUM ANALYZE machines`.

### DIFF
--- a/app/controllers/api/v1/groups/relationships/machines_controller.rb
+++ b/app/controllers/api/v1/groups/relationships/machines_controller.rb
@@ -10,7 +10,7 @@ module Api::V1::Groups::Relationships
     authorize :group
 
     def index
-      machines = apply_pagination(authorized_scope(apply_scopes(group.machines), with: Groups::MachinePolicy).preload(:product, :policy, :owner, license: %i[owner]))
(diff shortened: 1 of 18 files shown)


app/controllers/api/v1/groups/relationships/machines_controller.rb
```diff
@@ -10,7 +10,7 @@ class MachinesController < Api::V1::BaseController
     authorize :group
 
     def index
-      machines = apply_pagination(authorized_scope(apply_scopes(group.machines), with: Groups::MachinePolicy).preload(:product, :policy, :owner, license: %i[owner]))
+      machines = apply_pagination(authorized_scope(apply_scopes(group.machines), with: Groups::MachinePolicy).preload(:product, :policy, :owner, license: %i[policy owner]))
```
ezekg, line 13: Wish there was a way to tell active record that `machine.policy == license.policy`…

PR #944, opened 2025-04-29
Add note about lowercase format requirement for KEYGEN_ACCOUNT_ID in .env.sample

kliedo:
# Motivation

While setting up a self-hosted Keygen server, I found that using a UUID generated with `uuidgen` (which outputs uppercase by default) causes API requests to fail with an account is invalid error. Converting the UUID to lowercase resolves the issue.

To prevent this confusion for future users, I propose adding a note to the .env.sample file to clearly state that KEYGEN_ACCOUNT_ID must be in lowercase format.

## Changes
- Added a note in .env.sample to indicate that KEYGEN_ACCOUNT_ID must be in lowercase format.

## Impact

No breaking changes.
This is a documentation-only update to improve setup clarity for self-hosted users.


ezekg:
Thanks for the PR. I'll make a docs update mirroring this as well.

PR #948, opened 2025-05-08
Add Sidekiq plugin for bulk logs

### LINKED ISSUE #947: Bulk insert request and event logs
Instead of inserting 1 log at a time, let's batch these and bulk insert to reduce write load. Thinking the best way to do this would be via a Sidekiq plugin that handles queuing jobs and then combines them into a single job execution. This will be especially important when we start implementing entitlement meters and meter consumption in #400.

ezekg:
POC. Closes #947. Sidekiq plugin allows jobs to opt into bulk processing, where we batch multiple jobs into a single job. In our case, we use this for bulk inserting request and event logs, which during peak times can result in high write load.


spec/lib/perform_bulk_spec.rb
```diff
@@ -0,0 +1,362 @@
...
+  describe PerformBulk::BulkFetch do
+    let(:fetcher) { PerformBulk::BulkFetch }
+
+    it 'should bulk fetch from waiting queue' do
+      capsule.config[:bulk_batch_size] = 2
+      capsule.queues = [wait_queue.name]
+
+      # queue work
+      Sidekiq.redis do |conn|
+        conn.lpush("queue:#{wait_queue.name}", %w[foo bar baz])
```
ezekg, line 71: Could replace this with `wait_queue.push`?

spec/lib/perform_bulk_spec.rb
```diff
@@ -0,0 +1,362 @@
...
+
+  describe PerformBulk::UnitOfWork do
+    let(:unit_of_work) { PerformBulk::UnitOfWork.new(queue: "queue:#{wait_queue.name}", jobs: [1, 2, 3], config:) }
+
+    it 'responds quacks like Sidekiq::UnitOfWork' do
+      expect(unit_of_work).to respond_to :acknowledge
+      expect(unit_of_work).to respond_to :queue
+      expect(unit_of_work).to respond_to :queue_name
+      expect(unit_of_work).to respond_to :config
+      expect(unit_of_work).to respond_to :job
```
ezekg, line 127: Add `requeue`

app/workers/event_log_worker.rb
```diff
@@ -1,47 +1,12 @@
 # frozen_string_literal: true
 
+# FIXME(ezekg) need to move this to an EventLogWorker2 class to migrate
```
ezekg, line 3: Blocker

lib/perform_bulk.rb
```diff
@@ -0,0 +1,284 @@
...
+        end
+
+        logger.debug { "batching #{job_hashes.size} #{class_name.inspect} jobs" }
+
+        Runner.perform_async(class_name, args)
+      end
+    end
+  end
+
+  class Runner
```
ezekg, line 157: Can we make this so that it shows batch size in Sidekiq logs? e.g. `PerformBulk::Runner[42]: start|done`

lib/perform_bulk.rb
```diff
@@ -0,0 +1,284 @@
...
+        private :logger
+      end
+    end
+
+    private
+
+    def logger = Logger.instance
+  end
+
+  class Processor
```
ezekg, line 135: Can we make this so that it shows batch size in Sidekiq logs? e.g. `PerformBulk::Processor[100]: start|done`

ezekg: https://github.com/sidekiq/sidekiq/blob/b647e3adfe28be3bdefd90aabc345eeb46202591/lib/sidekiq/job_logger.rb#L30

app/services/broadcast_event_service.rb
```diff
@@ -52,17 +57,19 @@ def call
...
-            resource_type,
-            resource_id,
-            bearer_type,
-            bearer_id,
-            request_id,
-            idempotency_key,
-            metadata.to_json,
-            environment_id,
+          EventLogWorker2.perform_async(
+            'event_type_id' => event_type_id,
```
ezekg, line 61: Should set `id` to a UUIDv7

app/services/broadcast_event_service.rb
```diff
@@ -58,6 +58,7 @@ def call
             end
 
           EventLogWorker2.perform_async(
+            'id' => UUID7.generate,
```
ezekg, line 61: Could actually use `SecureRandom.uuid_v7` here

PR #949, opened 2025-05-11
feat: add dev-container for keygen + debian based-docker

aminya:
This adds a dev-container for Keygen, providing an isolated dev environment. 
Additionally, it adds a debian Debian-based Docker image, which is needed to get a better dev container experience (e.g. for docker-in-docker support).

I also separated the build stage into a build-base and build, in which the first one is independent of the gem files, making its caching better. This stage is used as the dev-container base.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -1,0 +1,39 @@
+{
+    "name": "keygen-api",
+    "build": {
+        "dockerfile": "debian.dockerfile",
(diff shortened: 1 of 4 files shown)


ezekg:
Thanks for the PR, but I don't want to support multiple Dockerfiles under the main repo.

Feel free to maintain this out-of-band.

PR #951, opened 2025-05-14
Add UUIDv7 as default record primary key

ezekg:
Should improve performance and risk in using batch operations like `find_each` using the default of sorting by primary key.

## Prereqs

- [ ] Talk to customers and make sure this won't be a breaking change, e.g. if a customer is expecting UUIDv4s, this would break that expectation — even if we have always reserved the right to change the format of new resource identifiers.

### DIFF
--- a/app/models/application_record.rb
+++ b/app/models/application_record.rb
@@ -48,7 +48,7 @@ class ApplicationRecord < ActiveRecord::Base
 
   private
 
-  def generate_uuid_v7_primary_key = self.id ||= UUID7.generate
+  def generate_uuid_v7_primary_key = self.id ||= SecureRandom.uuid_v7
 
   def self.uuid_primary_key? = attribute_types["id"].type == :uuid
   def uuid_primary_key?      = self.class.uuid_primary_key?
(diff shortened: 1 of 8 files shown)


app/models/application_record.rb
```diff
@@ -40,7 +44,12 @@ def self.sample = find_by(id: ids.sample) unless Rails.env.production?
...
   # This is a preventative measure to assert we never accidentally serialize
   # a model outside of our JSONAPI serializers
-  def serializable_hash(...)
-    raise NotImplementedError
-  end
+  def serializable_hash(...) = raise NotImplementedError
+
+  private
+
+  def generate_uuid_v7_primary_key = self.id ||= UUID7.generate
```
ezekg, line 51: Could use `SecureRandom.uuid_v7` introduced in Ruby 3.3

seuros:
You can use ulid, it cleaner and you can prefix it.
Here is a gem i build for it : https://github.com/seuros/activerecord-ulid

ezekg:
@seuros Our postgres primary keys are UUIDs, so that won't work. Changing primary key type is infeasible.

seuros:
In that case, yes.

PR #952, opened 2025-05-15
Add product authz scopes to webhooks

ezekg:
We've been relying on a broken authz model for webhooks i.r.t. products, where a product is allowed to create/delete endpoints and receive events that could be for resources outside of its jurisdiction — which could result in unauthorized spying if a product token is used outside of a trusted server-side environment. Though there's no signs of that occurring, this will likely become more of a risk with custom permissions recently being GA'd, so I'm tightening up the reigns preemptively.

## Prereqs

- [ ] Manually add `product_id` column to `webhook_endpoints` out-of-band. Then…
- [ ] Manually set `product_id` on all Zapier webhook endpoints, i.e. `webhook_endpoints.url like 'https://hooks.zapier.com/%`. In the case there are multiple products for a given account, we'll need to do a little bit of spelunking to determine correct product (likely looking at request logs from Zapier and peeking the requestor, which will be a specific product).

## Subreqs

- [ ] Verify Zapier integrations are still functioning as expected, i.e. events are still emitted to all Zapier endpoints, and no authz errors are occurring when Zapier handles them.

### DIFF
--- a/app/services/broadcast_webhook_service.rb
+++ b/app/services/broadcast_webhook_service.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+class BroadcastWebhookService < BaseService
+  include ActionPolicy::Behaviour
+
+  authorize :account
+  authorize :environment
+
+  def initialize(event:, account:, environment:, resource:, meta: nil)
+    @event       = event
+    @account     = account
+    @environment = environment
+    @resource    = resource
+    @meta        = meta
+  end
+
+  def call
+    selected_endpoint_ids = []
+
+    webhook_endpoints = account.webhook_endpoints.preload(product: { role: :permissions }).for_environment(
+      environment,
+      strict: true,
+    )
+
+    # skip resources that our endpoint's product aren't authorized to read (if any)
+    webhook_endpoints.find_each do |webhook_endpoint|
+      next unless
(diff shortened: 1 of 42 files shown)


app/services/broadcast_webhook_service.rb
```diff
@@ -0,0 +1,63 @@
...
+    # skip resources that our endpoint's product aren't authorized to read (if any)
+    webhook_endpoints.find_each do |endpoint|
+      action  = resource.class < Enumerable ? :index? : :show?
+      context = {
+        bearer: endpoint.product, # s/bearer/actor/
+        token: nil,
+      }
+
+      next unless
+        endpoint.product.nil? || allowed_to?(action, resource, context:)
```
ezekg, line 38: I want to push this to a background job, but it's tough because the resource may not exist anymore, and that would mean e.g. `*.deleted` webhooks events would be skipped for product webhook endpoints.

Maybe we can marshal it? idk

ezekg: Since most accounts that have a lot of endpoints have each endpoint subscribed to a small number of events, checking for event subscription before checking auth should save us from most inefficiency here. I'm going to leave as-is for simplicity's sake — we can always come back and tighten things up if performance is bad in production.

Premature-optimization and all that jazz.

PR #953, opened 2025-05-16
chore: fix Segmentation fault in tests

seuros:
YJIT and any_instance go together like Mentos and Coke.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,6 +56,7 @@ jobs:
       DATABASE_URL: postgres://postgres:postgres@localhost:5432
       REDIS_URL: redis://localhost:6379
       RAILS_ENV: test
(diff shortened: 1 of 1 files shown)


ezekg:
@seuros disabling YJIT makes tests less like production, which uses YJIT.

seuros:
YJIT needs load to get triggered, and when it does, it introduces random flaky behavior in the tests. That’s the core issue—mocking just doesn’t play well with any JIT, whether it’s YJIT, MJIT, or whatever-ZJIT comes next.

If you can figure out how to properly test without mocking or stubbing in this context, the problem disappears.

Your code is correct, I have seen this error many times before in my gems, which is exactly why I avoid mocks in tests now. 

You can’t safely mutate something that’s already been compiled.

If you want to keep that test and JIT, run it in isolation, it will never crash.

PR #957, opened 2025-05-22
Add Slack Connect

ezekg:
As part of onboarding and customer success, we'll send a Slack invite to new accounts that use a work email.

## Prereqs

- [x] Add `/chat` route to dashboard app. 

## Subreqs

- [ ] Add webhook to Slack bot (they expect a live URL).


app/models/concerns/slackable.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+module Slackable
+  extend ActiveSupport::Concern
+
+  included do
+    attr_accessor :skip_slack_invite
+
+    after_commit :send_slack_invite, on: :create,
```
ezekg, line 9: On the fence about this due to volume. Might be better if we add a trigger for this e.g. a "Connect Slack" button in the UI instead of full-send inviting every new signup to a private Slack channel. But maybe we can try it.

PR #995, opened 2025-09-26
Upgrade to Rails 8

ezekg:


spec/rails_helper.rb
```diff
@@ -204,10 +205,14 @@
...
+  # Load rake tasks once
+  mu = Once.new
+
   config.before type: :task do
-    require 'rake'
+    mu.synchronize do
+      require 'rake'
 
-    Rails.application.load_tasks
+      Rails.application.load_tasks
```
ezekg, line 215: Ref: https://github.com/rails/rails/issues/54002

PR #1011, opened 2025-10-23
fix: missing runtime dependencies (libexslt.so.0, liblzma.so.5) in dockerfile

fawwazid:


Dockerfile
```diff
@@ -43,27 +52,38 @@ RUN apk add --no-cache \
...
+  libxslt && \
   adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen
 
-COPY --from=build --chown=keygen:keygen \
-  /usr/local/bundle/ /usr/local/bundle
-
 WORKDIR /app
-COPY . /app
+
+# Copy Ruby gems from build stage
```
ezekg: Please remove all the superfluous comments.

Dockerfile
```diff
@@ -43,27 +52,38 @@ RUN apk add --no-cache \
...
+  xz-libs \
+  libxml2 \
+  libxslt && \
   adduser -h /app -g keygen -u 1000 -s /bin/bash -D keygen
 
-COPY --from=build --chown=keygen:keygen \
-  /usr/local/bundle/ /usr/local/bundle
-
 WORKDIR /app
-COPY . /app
```
ezekg: See https://github.com/keygen-sh/keygen-api/pull/1011/files#r2455318863

ezekg (changes requested):
Thanks for the PR. Let's try to keep the changeset as minimal as possible please.

PR #1034, opened 2026-01-08
Add dual-writes and read-your-own-writes libs

ezekg:
~~First pass. Commit history is a mess, but I'll clean it up later (probably).~~ Introduces a new Clickhouse database, which we'll eventually use as the primary source for logs and analytics, and soon, meter consumption ingest.

Introduces a new gem called `dual_writes` that replicates writes from Postgres to Clickhouse, which we're going to be using for request and event logs. Planning on running dual-writes for a few weeks — enough to collect a sufficient backlog in order to switch over analytical queries to Clickhouse without breaking everybody's charts. After the cutover, we can backfill historical log data for customers with longer retention periods than the default. Right now, the gem only has a Clickhouse strategy, but additional strategies can be implemented to e.g. replicate from Postgres to Postgres (e.g. for PS Metal).

In addition, we're also introducing a new gem called `read_your_own_writes`, or `ryow`, which lays the groundwork for a read replica database with a non-session-based "read-your-own-writes" resolver — since our API doesn't rely on sessions for most API requests (except Portal), clients are identified via a combination of account, session, auth, and IP.

Both of these gems will eventually be open sourced when time allows.

## Prereqs

- [x] Set up Clickhouse in prod via `CLICKHOUSE_DATABASE_URL`

## Later

- [ ] Set up read replica in prod via `REPLICA_DATABASE_CONNECTION_POOL_URL`

### DIFF
--- a/spec/lib/read_your_own_writes_spec.rb
+++ b/spec/lib/read_your_own_writes_spec.rb
@@ -103,14 +103,16 @@ describe ReadYourOwnWrites do
       end
 
       it 'should return timestamp after a write' do
-        write_request = build_request(path: '/v1/accounts/test/licenses/abc')
-        write_context = described_class.new(write_request)
-        write_context.update_last_write_timestamp
+        freeze_time do
+          write_request = build_request(path: '/v1/accounts/test/licenses/abc')
+          write_context = described_class.new(write_request)
+          write_context.update_last_write_timestamp
 
-        read_request = build_request(path: '/v1/accounts/test/licenses')
-        read_context = described_class.new(read_request)
+          read_request = build_request(path: '/v1/accounts/test/licenses')
+          read_context = described_class.new(read_request)
 
-        expect(read_context.last_write_timestamp).to be_within(1.second).of(Time.now)
+          expect(read_context.last_write_timestamp).to eq(Time.current)
+        end
       end
     end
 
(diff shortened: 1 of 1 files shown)


lib/dual_writes.rb
```diff
@@ -0,0 +1,458 @@
...
+        databases = Array(to)
+
+        raise ConfigurationError, 'to must be a symbol or array of symbols' unless
+          databases.all? { it.is_a?(Symbol) }
+
+        raise ConfigurationError, 'to cannot be empty' if
+          databases.empty?
+
+        # validate strategy can be looked up
+        Strategy.lookup(strategy)
```
ezekg: I don't like implicit validation like this. The strategy `lookup` method should return `nil`, and we can handle the `nil` case here by raising a `ConfigurationError` error.

lib/dual_writes.rb
```diff
@@ -0,0 +1,458 @@
...
+      end
+    end
+
+    def configuration = @configuration ||= Configuration.new
+    def configuration=(config)
+      @configuration = config
+    end
+
+    def configure
+      yield(configuration)
```
ezekg: Nit: remove parentheses

lib/read_your_own_writes.rb
```diff
@@ -0,0 +1,199 @@
...
+      def prefer_read_replica(**) = use_read_replica(**, always: false)
+    end
+
+    def with_primary_connection
+      ActiveRecord::Base.connected_to(role: :writing) { yield }
+    end
+
+    def with_read_replica_connection_unless_reading_own_writes
+      if ReadYourOwnWrites.reading_own_writes?(request)
+        yield # noop since already handled elsewhere
```
ezekg: Make it clearer that this is handled elsewhere via the default Rails database resolver

PR #1040, opened 2026-02-03
Update CI to use 32 CPU runners

ezekg:
Follow up to #1039.


ezekg:
The gains aren't there, unfortunately, so closing this for now. Seem to have some large Rspec specs, so might need to add `--group-by=example` to parallelize like we did for Cucumber, or find out why those specs are slow.

PR #1117, opened 2026-05-01
Add environment-scoped child sessions

ezekg:
Related to https://github.com/keygen-sh/keygen-portal/pull/181. Switching environments was broken in Portal because the environment session would overwrite the session for the nil environment, which broke switching back to nil. This separates the sessions.

### DIFF
--- a/app/controllers/api/v1/tokens_controller.rb
+++ b/app/controllers/api/v1/tokens_controller.rb
@@ -77,7 +77,9 @@ module Api::V1
       session = if request.origin == Keygen::Portal::ORIGIN
                   # TODO(ezekg) make default session expiry configurable
                   token.sessions.build(
-                    expiry: token.expiry.presence || 1.week.from_now,
+                    environment: token.environment,
+                    parent: current_session,
+                    expiry: token.expiry.presence || current_session&.expiry || 1.week.from_now,
                     user_agent: request.user_agent,
                     ip: request.remote_ip,
                   )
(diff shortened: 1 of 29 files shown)


app/models/concerns/environmental.rb
```diff
@@ -61,7 +61,7 @@ module Environmental
     #
     # Use :default to automatically configure a default environment for the model.
     # Accepts a proc that resolves into an Environment or environment ID.
-    def has_environment(default: nil, **kwargs)
+    def has_environment(default: nil, except: nil, **kwargs)
```
ezekg: Not sure I like this naming.

app/controllers/api/v1/tokens_controller.rb
```diff
@@ -77,7 +77,9 @@ def generate
       session = if request.origin == Keygen::Portal::ORIGIN
```
ezekg: If the bearer is different, we need to not generate a session, otherwise we'll clobber our existing session when e.g. generating a product token. This will also fail validation because the session bearer will differ.

app/models/token.rb
```diff
@@ -281,20 +282,22 @@ def regenerate!(session: nil, **)
...
-               )
-             end
-
-      generate!(**)
-
-      sesh
+      new_session = if session.present? && session.token == self
+                      sessions.build(
+                        environment: session.environment,
+                        parent: session.parent,
```
ezekg: Should be `session` not `session.parent`.

app/models/session.rb
```diff
@@ -41,6 +59,27 @@ class Session < ApplicationRecord
...
+    end
+  end
+
+  # assert that a child session cannot outlive its parent
+  validate on: %i[create update] do
+    next unless
+      parent.present? && parent.expiry? # anything goes if parent doesn't expire
+
+    if expiry.nil? || parent.expiry < expiry
+      errors.add :expiry, :invalid, message: 'cannot outlive parent session'
```
ezekg: Get rid of this

PR #1125, opened 2026-06-22
Upgrade Heroku stack

ezekg:


ezekg:
gg GitHub, merging did absolutely nothing 😄

Issue #1130, opened 2026-06-25
Support uploading custom account logos


Cazden:
Related to keygen-sh/keygen-portal#246. Portal renders a business logo in the sidebar combobox through [Logo.dev](https://www.logo.dev/) along with a monogram fallback. `Logo.dev` covers most established businesses/brands, but doesn't cover all cases:
* Air-gapped instances can't reach the img URLs provided by `Logo.dev`, so they only get the monogram.
* Self-hosted does have the option of configuring `Logo.dev` via environment, but it'd be more user-friendly to allow uploading a custom logo directly in the app.
* Anyone who wants exact branding/white-labeling is stuck with whatever `Logo.dev` returns (though they allow update requests).

PR #1142, opened 2026-07-29
Upgrade Rails

ezekg:
Re: https://github.com/rails/rails/security/advisories/GHSA-xr9x-r78c-5hrm. We aren't vulnerable, but since we plan on introducing image uploads via #1130 (and I was planning on using Active Storage), we should get ahead of this since it's a critical vulnerability.


Gemfile
```diff
@@ -4,6 +4,10 @@ source 'https://rubygems.org'
 ruby '3.4.9'
 
 gem 'rails', '~> 8.1.3.1'
+# FIXME(ezekg) zeitwerk >= 2.8 ignores do_not_eager_load() for collapsed directories,
+#              causing our clickhouse models to be eager loaded in environments
+#              without clickhouse (see config/initializers/zietwork.rb)
+gem 'zeitwerk', '2.7.3'
```
ezekg: https://github.com/fxn/zeitwerk/issues/337

